PR #1, opened 2015-08-15
Sps fixes

imbcmdth:
Roll up of my minor fixes to SPS parsing. No tests for `discardEmulationPreventionBytes` yet.

### DIFF
--- a/lib/exp-golomb.js
+++ b/lib/exp-golomb.js
@@ -53,7 +53,7 @@ muxjs.ExpGolomb = function(workingData) {
       workingBitsAvailable -= count;
     } else {
       count -= workingBitsAvailable;
-      skipBytes = count / 8;
+      skipBytes = Math.floor(count / 8);
 
       count -= (skipBytes * 8);
       workingBytesAvailable -= skipBytes;
--- a/test/test-sps.js
+++ b/test/test-sps.js
@@ -1,0 +1,25 @@
+window.testSPS = new Uint8Array([
+  0x00, 0x00, 0x00, 0x01,
+  103, 66, 192, 30,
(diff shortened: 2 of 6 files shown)


lib/exp-golomb.js
```diff
@@ -53,7 +53,7 @@ muxjs.ExpGolomb = function(workingData) {
       workingBitsAvailable -= count;
     } else {
       count -= workingBitsAvailable;
-      skipBytes = count / 8;
+      skipBytes = Math.floor(count / 8);
```
dmlap: Was this actually causing fractional numbers to be used in later calculations?

dmlap:
This one definitely could use a test case.

lib/exp-golomb.js
imbcmdth: Yes! But the issue was even more insidious.

The difficulty was that the error was transient because it only happened when `skipBits` crosses a "word" boundary (32-bit block.) What happens is that `skipBytes` is set to a float and `count` in the very next line (58) becomes `0`. 

While the first point means that `workingBytesAvailable` becomes a float, I am not sure that this was causing us any problems since we do not read all of the values in the RBSP.

But the second issue means that every time we crossed a word-boundary, we started reading again at the start of the word (because we never shifted to proper location after the boundary was crossed.)

imbcmdth:
Agreed.

test/test-sps.js
```diff
@@ -0,0 +1,25 @@
+window.testSPS = new Uint8Array([
```
dmlap: Given these are pretty tiny, I would consider just inlining them into the test case so you don't have go searching to find them. Not worth holding up the PR for, though.

PR #7, opened 2015-08-26
Start setting the baseMediaDecodeTime in fragments

imbcmdth:
We need to set the baseMediaDecodeTime so that MSE knows where, in time, it should plop segments

### DIFF
--- a/lib/transmuxer.js
+++ b/lib/transmuxer.js
@@ -464,6 +464,7 @@ AacStream = function() {
 
       // deliver the AAC frame
       this.trigger('data', {
+        pts: packet.pts,
         audioobjecttype: ((buffer[i + 1] >>> 6) & 0x03) + 1,
         channelcount: ((buffer[i + 1] & 1) << 3) |
           ((buffer[i + 2] & 0xc0) >>> 6),
@@ -493,6 +494,26 @@ AudioSegmentStream = function(track) {
   AudioSegmentStream.prototype.init.call(this);
 
   this.push = function(data) {
+    if (track.minPTS === undefined) {
+      track.minPTS = data.pts;
+    } else {
+      track.minPTS = Math.min(track.minPTS, data.pts);
+    }
+
+    if (track.maxPTS === undefined) {
+      track.maxPTS = data.pts;
+    } else {
+      track.maxPTS = Math.max(track.maxPTS, data.pts);
+    }
+
+    if (track && track.channelcount === undefined) {
+      track.audioobjecttype = data.audioobjecttype;
+      track.channelcount = data.channelcount;
+      track.samplerate = data.samplerate;
+      track.samplingfrequencyindex = data.samplingfrequencyindex;
+      track.samplesize = data.samplesize;
+    }
+
     // buffer audio data until end() is called
     aacFrames.push(data);
     aacFramesLength += data.data.byteLength;
@@ -902,10 +923,45 @@ VideoSegmentStream = function(track) {
   var
     sequenceNumber = 0,
     nalUnits = [],
-    nalUnitsLength = 0;
+    nalUnitsLength = 0,
+    config,
+    pps;
   VideoSegmentStream.prototype.init.call(this);
 
+  delete track.minPTS;
+
   this.push = function(data) {
+    if (track.minPTS === undefined) {
+      track.minPTS = data.pts;
+    } else {
+      track.minPTS = Math.min(track.minPTS, data.pts);
+    }
+
+    if (track.maxPTS === undefined) {
+      track.maxPTS = data.pts;
+    } else {
+      track.maxPTS = Math.max(track.maxPTS, data.pts);
+    }
+
+    // record the track config
+    if (data.nalUnitType === 'seq_parameter_set_rbsp' &&
+        !config) {
+      config = data.config;
+
+      track.width = config.width;
+      track.height = config.height;
+      track.sps = [data.data];
+      track.profileIdc = config.profileIdc;
+      track.levelIdc = config.levelIdc;
+      track.profileCompatibility = config.profileCompatibility;
+    }
+
+    if (data.nalUnitType === 'pic_parameter_set_rbsp' &&
+        !pps) {
+      pps = data.data;
+      track.pps = [data.data];
+    }
+
     // buffer video until end() is called
     nalUnits.push(data);
     nalUnitsLength += data.data.byteLength;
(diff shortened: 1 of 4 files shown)


dmlap:
@imbcmdth can you rebase this and see if that fixes the tests?

imbcmdth:
Tests Fixed!

lib/transmuxer.js
```diff
@@ -1029,61 +1142,27 @@ Transmuxer = function() {
...
+    .pipe(parseStream)
+    .pipe(elementaryStream);
+
   // remux the streams
   elementaryStream.pipe(aacStream);
   elementaryStream.pipe(h264Stream);
   elementaryStream.pipe(this.metadataStream);
 
-  // handle incoming data events
-  h264Stream.on('data', function(data) {
```
imbcmdth: This code has moved into the `push` function for the `VideoStream` constructor above. I saw no good reason for it to be here cluttering up the `Transmuxer` code.

lib/transmuxer.js
```diff
@@ -1029,61 +1142,27 @@ Transmuxer = function() {
...
-      videoTrack.profileCompatibility = config.profileCompatibility;
-    }
-    if (data.nalUnitType === 'pic_parameter_set_rbsp' &&
-        !pps) {
-      pps = data.data;
-      videoTrack.pps = [data.data];
-    }
-  });
-  // generate an init segment based on the first audio sample
-  aacStream.on('data', function(data) {
```
imbcmdth: This code has moved into the `push` function for the `AudioStream` constructor above. I saw no good reason for it to be here cluttering up the `Transmuxer` code.

lib/transmuxer.js
```diff
@@ -1029,61 +1142,27 @@ Transmuxer = function() {
...
-      audioTrack.audioobjecttype = data.audioobjecttype;
-      audioTrack.channelcount = data.channelcount;
-      audioTrack.samplerate = data.samplerate;
-      audioTrack.samplingfrequencyindex = data.samplingfrequencyindex;
-      audioTrack.samplesize = data.samplesize;
-    }
-  });
   // hook up the segment streams once track metadata is delivered
   elementaryStream.on('data', function(data) {
-    var triggerData = function(type) {
```
imbcmdth: I made this into a tiny little `Stream` so that `videoSegmentStream` and `audioSegmentStream` can simply be piped into it and it'll re-emit data through the `transmuxer`.

lib/transmuxer.js
```diff
@@ -1009,13 +1102,33 @@ Transmuxer = function() {
...
     audioTrack,
-    config,
-    pps,
+    earliestDTS,
 
     packetStream, parseStream, elementaryStream,
     aacStream, h264Stream,
     videoSegmentStream, audioSegmentStream;
 
+  this.dataEmitter = {
```
imbcmdth: Tiny `dataEmitter` stream to re-emit data it gets via `push` as the `Transmuxer`

lib/transmuxer.js
```diff
@@ -16,6 +16,9 @@ var TransportPacketStream, TransportParseStream, ElementaryStream,
     VideoSegmentStream, AudioSegmentStream, Transmuxer, AacStream,
     H264Stream, NalByteStream;
 
+// Helper functions
+var collectDTSInfo, clearDTSInfo, calculateTrackBMDT;
```
dmlap: DTS is a standard abbreviation so I think it's fine to use. I think "BMDT" is a bit more of a stretch and we should probably expand it.

lib/transmuxer.js
```diff
@@ -993,11 +1036,61 @@ VideoSegmentStream = function(track) {
...
+  if (track.minDTS === undefined) {
+    track.minDTS = data.dts;
+  } else {
+    track.minDTS = Math.min(track.minDTS, data.dts);
+  }
+
+  if (track.maxDTS === undefined) {
+    track.maxDTS = data.dts;
+  } else {
+    if (data.dts - track.maxDTS > 100) {
```
dmlap: This could use an explanatory comment.

lib/transmuxer.js
```diff
@@ -993,11 +1036,61 @@ VideoSegmentStream = function(track) {
...
+
+clearDTSInfo = function (track) {
+  delete track.minDTS;
+  delete track.maxDTS;
+  delete track.deltaDTS;
+};
+
+calculateTrackBMDT = function (track) {
+  var scale;
+  var nextExpectedBMDT = track.nextExpectedBMDT;
```
dmlap: One `var` per line unless you want to put together a patch to update the style across the whole project.

imbcmdth: Don't tempt me!

lib/transmuxer.js
dmlap: Very good move, it's much cleaner.

lib/transmuxer.js
dmlap: Is this intended to be accessed publicly? Seems like it should be scoped to the closure or at least underline-suffixed.

lib/transmuxer.js
```diff
@@ -1029,61 +1142,27 @@ Transmuxer = function() {
...
-        segment.set(mediaData, initSegment.byteLength);
-        self.trigger('data', {
-          type: type,
-          data: segment
-        });
-      };
-    }, i;
+    var i;
+
+    if (data.type === 'audio' || data.type === 'video') {
```
dmlap: Couldn't this be calculated by looking at the audio and video tracks whenever you need it?

imbcmdth: Not sure that I understand what you mean..

lib/transmuxer.js
```diff
@@ -993,11 +1036,61 @@ VideoSegmentStream = function(track) {
...
+
+calculateTrackBMDT = function (track) {
+  var scale;
+  var nextExpectedBMDT = track.nextExpectedBMDT;
+
+  track.baseMediaDecodeTime = track.minDTS - track.earliestDTS;
+
+  if (nextExpectedBMDT) {
+    // We arbitrarily decide that if the difference is greater than one
+    // second we should not attempt to "pull in" the segment in time
```
dmlap: I know this fixes some pretty big issues but I don't think we can use it in general. Unless I'm missing something, this will cause gradual errors in the relationship between the media timeline and PTS values. We need that relationship to stay stable so we can insert timed-metadata at the appropriate positions and seek accurately, for instance.

dmlap:
Spun off two issues we'll need to address at some point: #11 and #12. Otherwise, lgtm.

PR #13, opened 2015-09-04
608-over-708 caption parsing

dmlap:
This is still a work-in-progress-- it's not ready for a merge yet.

Add a caption stream to parse CEA-608 captions encapsulated in CEA-708 cc_data in an MPEG2-TS file. Currently, there is only very basic support for "pop-on" caption mode.


test/caption-stream-test.js
```diff
@@ -0,0 +1,283 @@
...
+
+  var captionStream;
+
+  module('Caption Stream', {
+    beforeEach: function() {
+      captionStream = new muxjs.mp2t.CaptionStream();
+    }
+  });
+
+  QUnit.skip('parses SEIs larger than 255 bytes', function() {
```
imbcmdth: The other ones are `not implemented` but this seems like a complete test. Just making sure that you didn't leave the skip in there by accident.

lib/caption-stream.js
```diff
@@ -0,0 +1,248 @@
...
+    this.displayed_ = '';
+    this.nonDisplayed_ = '';
+
+    this.push = function(packet) {
+      var data, swap, charCode;
+      // remove the parity bits
+      data = packet.ccData & 0x7f7f;
+
+      switch (data) {
+      case 0x0000:
```
imbcmdth: Would be nice to have these as named constants someplace. Would make this switch statement easier to grok.

lib/caption-stream.js
```diff
@@ -0,0 +1,248 @@
...
+
+    // only examine SEI NALs
+    if (event.nalUnitType !== 'sei_rbsp') {
+      return;
+    }
+    // parse the sei
+    sei = parseSei(event.data);
+
+    // ignore everything but user_data_registered_itu_t_t35
+    if (sei.payloadType !== 4) {
```
imbcmdth: Would be nice to have 4 be a named constant.

test/caption-stream-test.js
dmlap: Not quite finished-- I'd still have to fill out a more valid `data` property for the captionStream push a couple lines down.

imbcmdth:
Other than there being a few too many magic-numbers, LGTM :astonished:

lib/caption-stream.js
dmlap: I tried to comment each one and they're (currently) only used in this switch statement. Would you object to deferring that until we have to reference any of them elsewhere?

imbcmdth: My feeling is pretty much that you shouldn't need comments. But that is a stylistic preference. I would have a big object named `608PacketType` or somesuch with a nice name -> number mapping for the packets because I feel it makes it simpler to understand what it going on (considering it is already a not-straight forward state-machine type thing.)

test/caption-stream-test.js
imbcmdth: OK. Just checking.

lib/caption-stream.js
dmlap: So "nay" on the defer?

PR #14, opened 2015-09-08
Cross streams

imbcmdth:
First attempt at combining video/audio back into a single output segment.

Caveats:
* Stripped out the part where we try to pull in baseMediaDecodeTime of segments that are too far apart.

### DIFF
--- a/lib/transmuxer.js
+++ b/lib/transmuxer.js
@@ -14,7 +14,7 @@
 // object types
 var TransportPacketStream, TransportParseStream, ElementaryStream,
     VideoSegmentStream, AudioSegmentStream, Transmuxer, AacStream,
-    H264Stream, NalByteStream;
+    H264Stream, NalByteStream, CoalesceStream;
 
 // Helper functions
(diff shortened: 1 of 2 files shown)


lib/transmuxer.js
```diff
@@ -1114,6 +1103,75 @@ calculateTrackBaseMediaDecodeTime = function (track) {
...
+      offset += initSegment.byteLength;
+
+      // Append each moof (one per track) after the init segment
+      for (var i = 0; i < this.pendingBoxes.length; i++) {
+        segmentData.set(this.pendingBoxes[i], offset);
+        offset += this.pendingBoxes[i].byteLength;
+      }
+
+      // Emit the final segment via the transmuxer
+      transmuxer.trigger('data', {
```
dmlap: Could we have the transmuxer just listen for events on the coalesce stream and bubble them? Having this stream be aware of the transmuxer is inconsistent with the design of all the lower-level streams.

dmlap:
Tests are failing.

PR #15, opened 2015-09-08
Minor fixes

imbcmdth:
Two small fixes here:
* If a PMT has not been seen yet, defer processing PES packets until we have one.
* Clear the PPS and track config between segments so that we handle rendition switching correctly!

### DIFF
--- a/lib/transmuxer.js
+++ b/lib/transmuxer.js
@@ -114,6 +114,8 @@ TransportParseStream = function() {
   TransportParseStream.prototype.init.call(this);
   self = this;
 
+  this.packetsWaitingForPMT = [];
+  this.foundPMT = false;
   this.programMapTable = {};
 
   parsePsi = function(payload, psi) {
@@ -187,8 +189,14 @@ TransportParseStream = function() {
       offset += ((payload[offset + 3] & 0x0F) << 8 | payload[offset + 4]) + 5;
     }
 
+    self.foundPMT = true;
     // record the map on the packet as well
     pmt.programMapTable = self.programMapTable;
+
+    // if there are any packets waiting for a PMT to be found, process them now
+    while (self.packetsWaitingForPMT.length) {
+      self.processPES_.apply(self, self.packetsWaitingForPMT.shift());
+    }
   };
 
   parsePes = function(payload, pes) {
@@ -249,6 +257,7 @@ TransportParseStream = function() {
     var
       result = {},
       offset = 4;
+
     // make sure packet is aligned on a sync byte
     if (packet[0] !== 0x47) {
       return this.trigger('error', 'mis-aligned packet');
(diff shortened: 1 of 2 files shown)


lib/transmuxer.js
```diff
@@ -114,6 +114,8 @@ TransportParseStream = function() {
   TransportParseStream.prototype.init.call(this);
   self = this;
 
+  this.packetsWaitingForPMT = [];
```
dmlap: I've only been capitalizing the first letter of acronyms when camel-casing them elsewhere (e.g. `startPts`).

lib/transmuxer.js
```diff
@@ -187,8 +189,14 @@ TransportParseStream = function() {
       offset += ((payload[offset + 3] & 0x0F) << 8 | payload[offset + 4]) + 5;
     }
 
+    self.foundPMT = true;
```
dmlap: Is a separate variable for this necessary? Couldn't we just check `self.programMapTable !== undefined`?

dmlap:
The PPS and track config stuff could use a test.

lib/transmuxer.js
```diff
@@ -114,7 +114,8 @@ TransportParseStream = function() {
   TransportParseStream.prototype.init.call(this);
   self = this;
 
-  this.programMapTable = {};
+  this.packetsWaitingForPMT = [];
```
dmlap: `s/PMT/Pmt/`

lib/transmuxer.js
```diff
@@ -273,17 +280,26 @@ TransportParseStream = function() {
...
+      this.packetsWaitingForPMT.push([packet, offset, result]);
     } else {
-      result.streamType = this.programMapTable[result.pid];
-      result.type = 'pes';
-      parsePes(packet.subarray(offset), result);
+      this.processPES_(packet, offset, result);
     }
+  };
+
+  this.processPES_ = function (packet, offset, result) {
```
dmlap: Capitalization.

imbcmdth: So much nit-picky.

lib/transmuxer.js
```diff
@@ -189,6 +190,11 @@ TransportParseStream = function() {
 
     // record the map on the packet as well
     pmt.programMapTable = self.programMapTable;
+
+    // if there are any packets waiting for a PMT to be found, process them now
+    while (self.packetsWaitingForPMT.length) {
+      self.processPES_.apply(self, self.packetsWaitingForPMT.shift());
```
dmlap: Why is the `apply` necessary?

lib/transmuxer.js
dmlap: Stop being so sloppy ;)

PR #16, opened 2015-09-09
Collect captions in the coalesce stream before emitting

dmlap:
Captions require information about the start time of their associated segment in order to properly place them in the media timeline and that information is not available until after the entire segment is parsed. Buffer captions in the coalesce stream until the segment is ready and then calculate start and end times so that downstream consumers don't have to do it themselves.

### DIFF
--- a/lib/transmuxer.js
+++ b/lib/transmuxer.js
@@ -17,7 +17,7 @@ var TransportPacketStream, TransportParseStream, ElementaryStream,
     H264Stream, NalByteStream, CoalesceStream;
 
 // Helper functions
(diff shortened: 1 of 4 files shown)


lib/transmuxer.js
```diff
@@ -1055,6 +1055,11 @@ collectDTSInfo = function (track, data) {
   } else {
     track.minSegmentDTS = Math.min(track.minSegmentDTS, data.dts);
   }
+  if (track.minSegmentPts === undefined) {
```
imbcmdth: Might as well change the min/maxSegmentDTS names to min/maxSegmentDts while you are at it here just for consistency's sake.

imbcmdth:
LGTM

PR #17, opened 2015-09-10
Clean up the debug page

dmlap:
Remove some code duplication. Layout and style the page a bit differently so big videos don't break things and it's easier to see the form. Allow the "active" input to be selected without modifying the source of the page.

### DIFF
--- a/debug/css/main.css
+++ b/debug/css/main.css
@@ -160,6 +160,16 @@ nav a:hover {
 section {
   clear: both;
 }
+fieldset {
+  margin: 4px;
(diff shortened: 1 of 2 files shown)


imbcmdth:
Looks great. Love the radio boxes! We will need to be able to force the selection of a specific type of output eventually but that would require muxer changes to pass along some sort of options.

:palm_tree:

dmlap:
Agreed on all those points. I'm sure the next round of enhancements aren't far off.

PR #18, opened 2015-09-11
Add roll-up CEA-608 captions mode

dmlap:
Shift to a data model in Cea608Stream that more accurately simulates the theoretical captions decoder from the spec. Use the new model to implement roll-up captions mode with 2, 3, and 4 row support.

### DIFF
--- a/lib/caption-stream.js
+++ b/lib/caption-stream.js
@@ -177,17 +177,48 @@
   // semantics see
   // http://www.gpo.gov/fdsys/pkg/CFR-2010-title47-vol1/pdf/CFR-2010-title47-vol1-sec15-119.pdf
   var PADDING                    = 0x0000,
+
+      // Pop-on Mode
       RESUME_CAPTION_LOADING     = 0x1420,
+      END_OF_CAPTION             = 0x142f,
+
+      // Paint-on Mode
+      RESUME_DIRECT_CAPTIONING   = 0x1429,
+
+      // Roll-up Mode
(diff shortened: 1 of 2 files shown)


lib/caption-stream.js
```diff
@@ -218,6 +244,37 @@
         // start measuring the time to display the caption
         this.startPts_ = packet.pts;
         break;
+
+      case RESUME_DIRECT_CAPTIONING:
+        console.log('paint-on');
```
gkatsev: intentional?

lib/caption-stream.js
```diff
@@ -177,17 +177,48 @@
...
+      ERASE_NON_DISPLAYED_MEMORY = 0x142e;
+
+  // the index of the last row in a CEA-608 display buffer
+  var BOTTOM_ROW = 14;
+  // CEA-608 captions are rendered onto a 34x15 matrix of character
+  // cells. The "bottom" row is the last element in the outer array.
+  var createDisplayBuffer = function() {
+    var result = [], i = BOTTOM_ROW + 1;
+    while (i--) {
+      result.push('');
```
gkatsev: any particular reason this is an array of strings rather than an array of arrays?

lib/caption-stream.js
dmlap: Dang it.

lib/caption-stream.js
dmlap: As in, an array of arrays of single character strings?

PR #19, opened 2015-09-14
Support the backspace command code

dmlap:
In pop-on and roll-up modes, move the "cursor" back one cell and clear the character that is traversed when a backspace is received. We don't yet support any cursor-movement commands besides this one so the backspace is always applied to the last position in the base row for the current captioning mode.

### DIFF
--- a/test/caption-stream-test.js
+++ b/test/caption-stream-test.js
@@ -216,6 +216,48 @@
     }, 'parsed the third caption');
   });
 
+  test('backspaces are applied to non-displayed memory', function() {
+    var captions = [];
+    cea608Stream.on('data', function(caption) {
+      captions.push(caption);
+    });
+
+    // RCL, resume caption loading
+    cea608Stream.push({ ccData: 0x1420 });
+    // '01'
+    cea608Stream.push({
+      ccData: ('0'.charCodeAt(0) << 8) | '1'.charCodeAt(0)
+    });
+    // backspace
+    cea608Stream.push({ ccData: 0x1421 });
+    cea608Stream.push({
+      ccData: ('2'.charCodeAt(0) << 8) | '3'.charCodeAt(0)
+    });
+    // EOC, End of Caption
+    cea608Stream.push({ pts: 1 * 1000, ccData: 0x142f });
+    // EOC, End of Caption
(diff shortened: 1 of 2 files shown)


test/caption-stream-test.js
```diff
@@ -216,6 +216,48 @@
...
+    var captions = [];
+    cea608Stream.on('data', function(caption) {
+      captions.push(caption);
+    });
+
+    // RCL, resume caption loading
+    cea608Stream.push({ ccData: 0x1420 });
+    // '01'
+    cea608Stream.push({
+      ccData: ('0'.charCodeAt(0) << 8) | '1'.charCodeAt(0)
```
imbcmdth: This repeated snippet of code seems ripe for a function of some sort. Maybe with a short comment explaining it for us lay-people! :)

imbcmdth:
Otherwise, LGTM

test/caption-stream-test.js
dmlap: If I put the shift-ORing behind a function, you feel that would make the resulting test clearer? I thought this was sufficiently terse enough that it made it clear how `ccData` works without adding too much noise to the tests.

imbcmdth: I feel that having `('0'.charCodeAt(0) << 8) | '1'.charCodeAt(0)` repeated 4 times and `('2'.charCodeAt(0) << 8) | '3'.charCodeAt(0)` repeated twice is silly. At minimum, we should define "BACKSPACE_CCDATA" and "TESTCHAR_CCDATA" or something similar.

dmlap: I do not see what additional information this:
```js
cea608Stream.push({
  ccData: BACKSPACE_CCDATA
});
```
provides over this:
```js
cea608Stream.push({
  ccData: 0x1421 // backspace
});
```
(in fact, it is a little more opaque than the latter example).

In my mind, one duty of the tests is to teach the interface to the code to someone approaching the project for the first time. Using constants over literals hides some important details of actual operation. Is your objection mostly aesthetic?

imbcmdth: My object was to having code such as `('0'.charCodeAt(0) << 8) | '1'.charCodeAt(0)` repeated multiple times. Doing the change you show above is adequate.

PR #20, opened 2015-09-14
Selective output

imbcmdth:
Transmuxer debugging page now has changes that allow for greater debugging options by enabling the user to force the transmuxer to output only audio or only video segments. Helps you narrow down which of the two segment types is causing a problem with MSE.

Required changes the transmuxer to support non-remuxed output. Added tests to verify that the output from the transmuxer can be specified with an options-object at creation.

### DIFF
--- a/debug/index.html
+++ b/debug/index.html
@@ -14,6 +14,12 @@
   <link rel="stylesheet" href="css/main.css">
 
   <script src="js/vendor/modernizr-2.6.2.min.js"></script>
+  <style>
+    fieldset {
+      padding: 10px;
+      border: 1px solid #ccc;
+    }
+  </style>
 </head>
 <body>
   <!--[if lt IE 7]>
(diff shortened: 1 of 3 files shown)


debug/index.html
```diff
@@ -52,13 +58,28 @@ <h1 class="title">Transmux Analyzer</h1>
...
+              <legend>M2TS Input</legend>
+              <div>
+                <input id="original-active" type=radio name=active checked value="original">
+                <label>
+                  Your original MP2T segment:
+                  <input type="file" id="original">
+                </label>
+              </div>
+              <div>
+                <label><input id="combined-output" type=checkbox name=combined checked value="combined">&nbsp;Remux output into a single output?
```
dmlap: It doesn't really matter since this is only a test page but `&nbsp;`?

imbcmdth: It's only a test page. I needed a little spacing because it was too close... ::deal_with_it.gif:: :smile:

dmlap:
:guitar:

PR #22, opened 2015-09-17
Reset bmdt

imbcmdth:
A laundry list of fixes in the PR (in roughly top-down order):
* Rewrote transport packet parsing to be more robust when confronted with strange segmentation strategies - again!
* On a ~~BMDT reset (PTS - discontinuity)~~ the first segment, we now remove AAC frames that extend before the first PTS of the video track.
* Extended tracking previously done with `timelineStartDts` to track both audio and video tracks and both dts and pts values for each.
* Caption `startTime` and `endTime` are now relative to the last time base media decode time was set to zero.

### DIFF
--- a/lib/transmuxer.js
+++ b/lib/transmuxer.js
@@ -54,24 +54,102 @@ mp4 = muxjs.mp4;
  * Splits an incoming stream of binary data into MPEG-2 Transport
  * Stream packets.
  */
+
 TransportPacketStream = function() {
   var
     buffer = new Uint8Array(MP2T_PACKET_LENGTH),
-    end = 0;
+    bytesInBuffer = 0;
 
   TransportPacketStream.prototype.init.call(this);
 
-  /**
-   * Deliver new bytes to the stream.
-   */
-  this.push = function(bytes) {
-    var i = 0;
+   // Deliver new bytes to the stream.
 
-    // parse out all the completed packets
-    do {
+  this.push = function(bytes) {
+    var
+      bytesRemaining,
+      i = 0;
+
+    bytes = this.makePacketFromBufferedData_(bytes);
+
+    // If bytes is undefined, then we don't have enough data for one
+    // complete packet so just return and wait for more
+    if (!bytes) {
+      return;
+    }
+
+    bytesRemaining = bytes.byteLength;
+
+    // Parse out all the completed packets
+    while (i + MP2T_PACKET_LENGTH <= bytes.byteLength) {
       this.trigger('data', bytes.subarray(i, i + MP2T_PACKET_LENGTH));
       i += MP2T_PACKET_LENGTH;
-    } while (i + MP2T_PACKET_LENGTH < bytes.byteLength);
+      bytesRemaining -= MP2T_PACKET_LENGTH;
+    }
+
+    // Buffer any partial packets left over
+    if (bytesRemaining > 0) {
+      buffer.set(bytes.subarray(i));
+      bytesInBuffer = bytesRemaining;
+    }
+  };
+
+  // Attempt to complete any partial packets from data that is in the
+  // `buffer` and new data we've just recieved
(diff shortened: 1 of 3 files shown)


lib/transmuxer.js
```diff
@@ -54,52 +54,101 @@ mp4 = muxjs.mp4;
...
+
+    while (true) {
+      // If bytes starts with a sync-byte assume the `buffer` is junk
+      // and don't change a thing because we are starting with an
+      // entire packet in the new data
+      if (bytes[0] === 0x47) {
+        // Clear the buffer
+        bytesInBuffer = 0;
+        return bytes;
+      } else {
```
dmlap: This `else` is unnecessary.

lib/transmuxer.js
```diff
@@ -54,52 +54,101 @@ mp4 = muxjs.mp4;
...
+        while (i < bytesInBuffer && buffer[i] !== 0x47) {
+          i++;
+        }
+
+        if (i === bytesInBuffer) {
+          // We never found a sync-byte so throw out all the data and try
+          // again
+          bytesInBuffer = 0;
+          continue;
+        } else if (i > 0) {
```
dmlap: This `else` nesting is also unnecessary.

lib/transmuxer.js
```diff
@@ -54,52 +54,101 @@ mp4 = muxjs.mp4;
...
+      bytesInBuffer = bytesRemaining;
+    }
+  };
+
+  // Attempt to complete any partial packets from data that is in the
+  // `buffer` and new data we've just recieved
+  this.makePacketFromBufferedData_ = function (bytes) {
+    var bytesNeeded, bytesLeft;
+
+    while (true) {
```
dmlap: Discussed offline: @imbcmdth and I agreed we need to try to take a slightly different tack than this implementation.

lib/transmuxer.js
```diff
@@ -1333,19 +1391,32 @@ Transmuxer = function(options) {
...
       videoSegmentStream.flush();
     }
 
     if (audioSegmentStream) {
-      audioTrack.timelineStartDts = timelineStartDts;
+      // Use the video's earliest Dts value unless this is an audio-only stream
+      audioTrack.timelineStartInfo = timelineStartInfo['video'] || timelineStartInfo['audio'];
+
+      // When we are resetting the base media decode time due to a
+      // discontinuity, we need to remove any audio frames that are
```
dmlap: An explanation of why we need to remove the audio frames would be nice to add.

lib/transmuxer.js
```diff
@@ -1250,7 +1283,19 @@ Transmuxer = function(options) {
...
+    // we were when the base media decode time was last (re)set to zero
+    timelineStartInfo = {},
+
+    // Signal that we are in the process of resetting the base media
+    // decode time.
+    // One effect of this state is that we trim audio frames that start
+    // earlier than the video track so that audio always has an initial
+    // base media decode time that is equal to or greater than the video
+    // track.
+    resettingBaseMediaDecodeTime = true,
```
dmlap: I have what may be an irrational dislike for boolean state variables. It may be because once you get started with them, it's so easy to add another without considering the additional "modes" you have created for an object. Are we only resetting the base decode when seeking or playing over a discontinuity? Could we simply discard and recreate the Transmuxer stack in that case?

lib/transmuxer.js
```diff
@@ -54,53 +54,83 @@ mp4 = muxjs.mp4;
  * Splits an incoming stream of binary data into MPEG-2 Transport
  * Stream packets.
  */
+
 TransportPacketStream = function() {
   var
+    SYNC_BYTE = 0x47,
```
dmlap: This should be moved out of the constructor.

imbcmdth: Seriously?

PR #24, opened 2015-09-25
Save and restore the state of inputs in the debugger across page loads..

imbcmdth:
..because changing the codecs string every time is annoying!


dmlap:
I just had a PR queued up that would switch audio back to vanilla AAC but this is nicer. LGTM

PR #25, opened 2015-09-26
Fix everything and emit ID3 metadata

imbcmdth:
* Fix NAL unit parsing when the NAL units begin with the 3-byte start code
* Fix ADTS parsing:
 * When the ADTS packet crosses a PES boundary
 * When the ADTS byte-stream becomes desynchronized
* Listen for and emit ID3 metadata in the coalesce stream with each segment
* Tests!

### DIFF
--- a/test/transmuxer-test.js
+++ b/test/transmuxer-test.js
@@ -1710,10 +1710,16 @@ validateTrack = function(track, metadata) {
   equal(mdia.boxes[2].type, 'minf', 'wrote the media info');
 };
 
-validateTrackFragment = function(track, segment, metadata) {
+validateTrackFragment = function(track, segment, metadata, type) {
   var tfhd, trun, sdtp, i, j, sample, nalUnitType;
   equal(track.type, 'traf', 'wrote a track fragment');
-  equal(track.boxes.length, 4, 'wrote four track fragment children');
+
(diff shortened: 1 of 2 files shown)


test/transmuxer-test.js
```diff
@@ -1865,4 +1965,134 @@ test('can be reused for multiple TS segments', function() {
...
+      0x00, 0x00, 0x00 // end code
+    ])
+  });
+
+  equal(nalUnits.length, 2, 'found two nals');
+  deepEqual(nalUnits[0], new Uint8Array([0x09, 0xFF]), 'has the proper payload');
+  deepEqual(nalUnits[1], new Uint8Array([0x12, 0xDD]), 'has the proper payload');
+});
+
+test('parses nal units surrounded by an unreasonable amount of zero-bytes', function(){
```
dmlap: Funny guy.

test/transmuxer-test.js
```diff
@@ -1554,9 +1631,6 @@ test('can specify that we want to generate separate audio and video segments', f
   equal('mdat', boxes[3].type, 'generated a mdat box');
 });
 
-// not handled: ADTS with CRC
```
dmlap: Do we have a test case with a CRC?

PR #26, opened 2015-09-28
Ignore the "mpeg version" bit in the ADTS header when looking for the syncword

imbcmdth:
AAC is defined in both mpeg-2 and mpeg-4 specs and I don't think this distinction matters for our purposes.

### DIFF
--- a/test/transmuxer-test.js
+++ b/test/transmuxer-test.js
@@ -1535,6 +1535,29 @@ test('resyncs data in aac frames that contain garbage', function() {
             'extracted the second AAC frame');
 });
 
+test('ignores audio "MPEG version" bit in adts header', function() {
+  var frames = [];
+  aacStream.on('data', function(frame) {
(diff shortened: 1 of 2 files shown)


test/transmuxer-test.js
```diff
@@ -1535,6 +1535,29 @@ test('resyncs data in aac frames that contain garbage', function() {
...
 
+test('ignores audio "MPEG version" bit in adts header', function() {
+  var frames = [];
+  aacStream.on('data', function(frame) {
+    frames.push(frame);
+  });
+  aacStream.push({
+    type: 'audio',
+    data: new Uint8Array([
+      0xff, 0xf8,       // MPEG-2 audio, no CRC
```
dmlap: Looks like this packet *does* have a CRC, right?

dmlap:
Other than the comment on the test, LGTM

PR #27, opened 2015-10-05
First working merger of the MSE M2TS parsing front-end with the FLV generating backend

imbcmdth:
Combined flash and mp4 generating front-ends so that there is only one M2TS decoding path. Major project reorganization.

* Moved most of the \*Streams in `transmuxer.js` to `codecs/*` and `mp2ts/*` as they are now shared between both mp4 and flv generators
* Changed `SegmentParser` into a stream and renamed it to `flv/transmuxer.js`
* Moved `transmuxer.js` to `mp4/transmuxer.js`
* Removed all the legacy cruft

The "exported" object hierarchy mirrors the directory structure and now looks like:
```javascript
muxjs.codecs.AacStream
muxjs.codecs.H264Stream

muxjs.mp2ts.TransportPacketStream
muxjs.mp2ts.TransportParseStream
muxjs.mp2ts.ElementaryStream
muxjs.mp2t.MetadataStream 
muxjs.mp2t.CaptionStream 

muxjs.mp4.Transmuxer
muxjs.mp4.ftyp
muxjs.mp4.mdat
muxjs.mp4.moof
muxjs.mp4.moov
muxjs.mp4.initSegment

muxjs.flv.Transmuxer
muxjs.flv.FlvTag

muxjs.utils.ExpGolomb
muxjs.utils.Stream

muxjs.tools.inspectMp4
muxjs.tools.inspectorFlv
```


legacy/segment-parser.js
```diff
@@ -1,497 +1,568 @@
...
+
+  return tag;
+};
+
+/**
+ * Constructs a single-track, media segment from AAC data
+ * events. The output of this stream can be fed to flash.
+ */
+// TODO: share common code with VideoSegmentStream
+AudioSegmentStream = function(track) {
```
dmlap: I'm correct to assume we'll share the implementation before merging, right?

lib/transmuxer_common.js
```diff
@@ -0,0 +1,944 @@
+/**
+ * mux.js
```
dmlap: How would you feel about breaking this up by format? "common" isn't bad but `mp2t`, `h264`, and `aac` might be better descriptions of the content.

debug/legacy.html
```diff
@@ -103,19 +115,244 @@ <h1 class="title">Transmux Analyzer</h1>
...
+              case 2:
+                arr.push(hex(data[i++]));
+              case 1:
+                arr.push(hex(data[i++]));
+            }
+            data = data.subarray(i);
+          }
+          return arr.join(' ');
+        },
+        parseAVCTag = function (tag, obj) {
```
dmlap: We *could* pull this out into a first class FLV inspector utility...

lib/flv/transmuxer.js
```diff
@@ -0,0 +1,563 @@
...
+    // Continue with the flush process now
+    this.trigger('done');
+  };
+};
+
+VideoSegmentStream.prototype = new muxjs.Stream();
+
+/**
+ * A Stream that can combine multiple streams (ie. audio & video)
+ * into a single output segment for MSE. Also supports audio-only
```
dmlap: > for MSE

Is that true?

lib/flv/transmuxer.js
```diff
@@ -0,0 +1,563 @@
...
+    if (output.track.type === 'audio') {
+      this.audioTrack = output.track;
+      this.audioTags = output.tags;
+      this.pendingTracks++;
+    }
+  };
+};
+
+CoalesceStream.prototype = new muxjs.Stream();
+CoalesceStream.prototype.flush = function() {
```
dmlap: Not something we necessarily have to fix in this PR but we have a ton of duplication in the "remuxer" streams across MP4 and FLV.

npm-debug.log
```diff
@@ -0,0 +1,74 @@
+0 info it worked if it ends with ok
```
dmlap: Accidentally committed, right?

dmlap:
Would be nice to update the one comment I noted and drop npm-debug.log from the changeset. Otherwise LGTM

debug/legacy.html
imbcmdth: Way ahead of you :stuck_out_tongue:

npm-debug.log
imbcmdth: Yup.

lib/flv/transmuxer.js
imbcmdth: I agree though it would involve some more refactoring. Coalesce is similar but not the same. Same with Video/AudioSegmentStream. It would be nice if more of that code were in reusable functions so that we had a finer reuse granularity than simply "stream." That should be something we should work towards when we have the time.

PR #28, opened 2015-10-21
Fix PES Parsing

imbcmdth:
* Moved parsePes to the ElementaryStream *after* the entire packet has been received
* Fixed a minor bug that resulted in PTS/DTS values potentially being off by 1
* Fixed/moved tests since a major portion of the TransportParseStream's work is now done in ElementaryStream
* Added a case which tests header reassembly across packets

### DIFF
--- a/test/transmuxer-test.js
+++ b/test/transmuxer-test.js
@@ -32,6 +32,7 @@ var
   validateTrack,
   validateTrackFragment,
 
+  pesHeader,
   transportPacket,
   videoPes,
   audioPes,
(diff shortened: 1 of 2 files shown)


test/transmuxer-test.js
```diff
@@ -668,19 +574,22 @@ test('parses standalone program stream packets', function() {
...
   });
   elementaryStream.flush();
 
-  equal(1, packets.length, 'built one packet');
-  equal('audio', packets[0].type, 'identified audio data');
-  equal(19, packets[0].data.byteLength, 'parsed the correct payload size');
+  equal(packets.length, 1, 'built one packet');
+  equal(packets[0].type, 'audio', 'identified audio data');
+  equal(packets[0].data.byteLength, packetData.length, 'parsed the correct payload size');
+  equal(packets[0].pts, 7, 'correctly parsed the pts value');
```
dmlap: Nice cleanup.

dmlap:
We may have to handle split headers for the PAT and PMT too but we can tackle that later if necessary. LGTM.

PR #29, opened 2015-10-26
More debugging tools

dmlap:
Split pane diff view and a new multi-segment player test page.

### DIFF
--- a/debug/js/main.js
+++ b/debug/js/main.js
@@ -1,1 +1,23 @@
+(function(muxjs) {
+  'use strict';
 
+  var diffParsed = function(expected, actual) {
+    var expectedLines = muxjs.tools.textifyMp4(expected, null, ' ').split('\n');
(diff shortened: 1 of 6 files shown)


debug/js/main.js
```diff
@@ -1 +1,65 @@
...
+      video.addEventListener('error', logSourceBuffer);
+      video.addEventListener('error', function() {
+        video.classList.add('error');
+      });
+
+      return callback(mediaSource, sourceBuffer);
+    });
+  }
+
+  muxjs.debug = {
```
imbcmdth: I don't like that these helper functions are in the muxjs "namespace".

I'm even second guessing our inclusion of mp4-inspector and flv-inspector in the namespace since they are only used in these debugging pages and are not even part of what would be the mux.js "build".

dmlap: It doesn't really matter where they end up to me. Do you have another suggestion?

PR #30, opened 2015-10-26
Browserify-ify

imbcmdth:
Started using browserify + babelify to build the project and added jshint step while I was at it for good measure. Everything is a nice neat module and dependency management is handled implicitly now. 

Could still do with some more modularization (for example, the m2ts.js file does way too much.) 

!!DO NOT MERGE!!

..NEEDS TESTS UPDATED..

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,3 @@
 .DS_Store
 /node_modules/
 *~
-dist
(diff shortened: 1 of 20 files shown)


dmlap:
Obsoleted by #44.

PR #32, opened 2015-10-30
elegantly handle malformed ID3 tags

saxena-gaurav:


test/metadata-stream-test.js
```diff
@@ -269,7 +269,8 @@
       trackId: 7,
       pts: 1000,
       dts: 900,
-      data: front
+      data: front,
+      dataAlignmentIndicator: 0x01
```
imbcmdth: By the time these packets exist, `dataAlignmentIndicator` is either `true` or `false` - it was coerced into a boolean from the value in the PES header.

imbcmdth:
Otherwise LGTM

PR #33, opened 2015-11-05
H264frame fix

imbcmdth:
There are cases when nalUnit streams may not start with an AUD and this can result in `h264Frame` being undefined as we parse more

### DIFF
--- a/lib/flv/transmuxer.js
+++ b/lib/flv/transmuxer.js
@@ -185,7 +185,6 @@ VideoSegmentStream = function(track) {
     nalUnits = [],
     nalUnitsLength = 0,
     config,
-    pps,
     h264Frame;
   VideoSegmentStream.prototype.init.call(this);
 
@@ -221,12 +220,6 @@ VideoSegmentStream = function(track) {
       currentNal,
       tags = [];
 
-    // return early if no video data has been observed
-    if (nalUnits.length === 0) {
-      this.trigger('done');
-      return;
-    }
-
     // Throw away nalUnits at the start of the byte stream until we find
     // the first AUD
     while (nalUnits.length) {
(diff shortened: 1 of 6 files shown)


lib/flv/transmuxer.js
```diff
@@ -227,26 +227,35 @@ VideoSegmentStream = function(track) {
...
+        config = currentNal.config;
+        track.width = config.width;
+        track.height = config.height;
+        track.sps = [currentNal.data];
+        track.profileIdc = config.profileIdc;
+        track.levelIdc = config.levelIdc;
+        track.profileCompatibility = config.profileCompatibility;
+        h264Frame.endNalUnit();
+      } else if (currentNal.nalUnitType === 'pic_parameter_set_rbsp') {
+        track.newMetadata = true;
```
dmlap: Tangential to this review but why is this boolean necessary? Wouldn't `track.pps.length` suffice?

lib/flv/transmuxer.js
```diff
@@ -227,26 +227,35 @@ VideoSegmentStream = function(track) {
       return;
     }
 
+    // Throw away nalUnits at the start of the byte stream until we find
+    // the first AUD
```
dmlap: This step is not necessary for the MP4 transmuxer?

dmlap:
A couple questions but possibly no action necessary. LGTM

lib/flv/transmuxer.js
imbcmdth: Unfortunately, `track.[pps/sps]` is simply set every time we see a pps or sps so except for the first time it'll always be 1. We need to emit the metadata tags for based on sps and pps every keyframe so we need to keep the data around.

The boolean just signals that the metadata *may* have changed so emit it even if we *aren't* at a keyframe.

lib/flv/transmuxer.js
imbcmdth: I can't think of a good reason NOT to do this..

PR #36, opened 2015-11-17
Re-order accumulated captions by PTS before flushing

willkelleher:
The Cea608 caption parsing currently breaks if the caption packets arrive with out-of-order PTS values.  Not sure if this is the best remedy, but it works.

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -200,7 +200,7 @@
   var createDisplayBuffer = function() {
     var result = [], i = BOTTOM_ROW + 1;
     while (i--) {
-      result.push('');
+      result.push({});
     }
     return result;
   };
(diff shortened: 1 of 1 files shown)


willkelleher:
Looking at test failures now...

Anyone know if there is a way to enforce presentation order on the caption packets before it enters Cea608 processing?  The roll-up display mode isn't very well suited to this sort of re-ordering.

dmlap:
Woah, thanks for finding this! If you're having trouble with the tests, feel free to ask here.

A couple nitpicky style notes:
- we use camelCase for variable names, not snake_case
- `{`s go on the same line as the statement that opened them, not standalone on the next line
- use `===`, not `==`

We should definitely have jshint setup to check for this already, sorry. Would you mind making those adjustments?

willkelleher:
No problem! Let me clean this up a bit and get all the tests to pass. I
will let you know if I have any more specific questions.

willkelleher:
Tests are passing, style issues should be fixed.  I'm still not convinced that this is the best solution to the problem...so if anyone has other ideas, please chime in.

dmlap:
I spent some time thinking about this one and I believe we'd still have a problem with this solution. Basically, we can't begin "interpreting" the 608 data until we're sure no new packet will arrive between the PTS values we collected so far. That means some sort of signaling mechanism to flush the 608 parser whenever we hit a new GOP which is a little different from what's happening in this patch.

We've been talking about fragmenting the transmuxer output at GOP boundaries instead of per-input, like we're doing today. If we started doing that, I think it would be a lot easier to signal the 608 parser to flush its input at that time too.

willkelleher:
I see what you mean, but I don't think fragmenting at GOP boundaries would help much, since those could be nearly the size of an input segment.  How would this work with a GOP of 270?

dmlap:
I think we may have solved this via #42. Do you still see this problem in master now?

imbcmdth:
This should have been corrected in #42. Thanks for reporting the issue! If you still have issues with the current master don't hesitate to open another issue.

PR #38, opened 2015-11-17
Fix for large PTS/DTS values resulting in negative values

imbcmdth:
The basic change is to construct only the first 31-bits of the 33bit pts/dts values with bitwise ops since bitwise operations always work on *signed* 32bit integers. The remaining two bits are then added via math-wise operations.


dmlap:
LGTM

PR #39, opened 2015-11-17
Allow a base-baseMediaDecodeTime to specified when the transmuxer is constructed

imbcmdth:
Allow specifying a time (in 90khz) to consider as the `baseMediaDecodeTime` of the first segment pushed through the transmuxer. Subsequent segments will be placed relative to the first in time.

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -254,7 +254,7 @@ TransportParseStream.STREAM_TYPES  = {
  * packets plus relevant metadata that has been extracted from the
  * container.
  */
-ElementaryStream = function(timelineBaseDts) {
+ElementaryStream = function() {
   var
     // PES packet fragments
     video = {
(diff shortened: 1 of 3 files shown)


lib/m2ts/m2ts.js
```diff
@@ -402,7 +404,9 @@ ElementaryStream = function() {
         for (k in programMapTable) {
           if (programMapTable.hasOwnProperty(k)) {
             track = {
-              timelineStartInfo: {}
+              timelineStartInfo: {
+                baseDts: timelineBaseDts
```
dmlap: Should `baseDts` just be called `baseMediaDecodeTime` for consistency?

lib/m2ts/m2ts.js
```diff
@@ -254,7 +254,7 @@ TransportParseStream.STREAM_TYPES  = {
  * packets plus relevant metadata that has been extracted from the
  * container.
  */
-ElementaryStream = function() {
+ElementaryStream = function(timelineBaseDts) {
```
dmlap: Naming nitpick: I think we should stick with MSE terminology where it makes sense and this seems like a good candidate. `timelineOffset`?

dmlap:
Naming questions, otherwise LGTM

PR #40, opened 2015-11-23
Do not output ID3 and caption times based on baseMediaDecodeTime

imbcmdth:
Instead of outputting ID3 and captions in absolutely time, go back to placing them relative to the start of the *first* segment since we last set the `baseMediaDecodeTime`. Let `videojs-contrib-media-sources` do the translation in time based on `timestampOffset` which follows what we must do in the Flash case.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -418,7 +418,6 @@ CoalesceStream.prototype.flush = function() {
     id3,
     initSegment,
     timelineStartPts = 0,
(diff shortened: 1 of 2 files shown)


dmlap:
Neither @imbcmdth or I are totally thrilled with the asymmetry between the FLV-outputs and MP4-outputs but we can't think of a simple way to unify them off-hand. For the moment, upstream users can adjust their timed metadata placement for both FLV and MP4 and at least be consistent.

PR #42, opened 2015-12-04
Caption ID3 Parsing Improvements

imbcmdth:
## Fix sei/caption parsing
* SEI nal units are purged of any emulation prevention bytes
* SEI parsing now works with SEI nals containing multiple messages
* Caption parsing now treats preamble address codes as spaces, improving readability
* Caption parsing now correctly ignores redundant control codes
* Captions packets are reordered based on PTS before being processed to handle out-of-order caption packets arising from B-frames

## Fix parsing of the size fields of ID3 tags
* All sizes or lengths in ID3 tags and frames are in "sync-safe integer" format - parse them correctly
* Fix tests which were not creating the correct tags

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -20,38 +20,62 @@
   // payload type field to indicate how they are to be
   // interpreted. CEAS-708 caption content is always transmitted with
   // payload type 0x04.
-  var USER_DATA_REGISTERED_ITU_T_T35 = 4;
+  var USER_DATA_REGISTERED_ITU_T_T35 = 4,
+      RBSP_TRAILING_BITS = 128;
 
   /**
    * Parse a supplemental enhancement information (SEI) NAL unit.
+   * Stops parsing once a message of type ITU T T35 has been found.
    *
    * @param bytes {Uint8Array} the bytes of a SEI NAL unit
    * @return {object} the parsed SEI payload
    * @see Rec. ITU-T H.264, 7.3.2.3.1
    */
   var parseSei = function(bytes) {
-    var result = {
-      payloadType: -1,
-      payloadSize: 0,
-    }, i;
+    var
+      i = 0,
+      result = {
+        payloadType: -1,
+        payloadSize: 0,
+      },
+      payloadType = 0,
+      payloadSize = 0;
 
-    // parse the payload type
-    // if the payload type is not user_data_registered_itu_t_t35,
-    // don't bother parsing any further
-    if (bytes[1] !== USER_DATA_REGISTERED_ITU_T_T35) {
-      return result;
+    // go through the sei_rbsp parsing each each individual sei_message
+    while (i < bytes.byteLength) {
+      // stop once we have hit the end of the sei_rbsp
+      if (bytes[i] === RBSP_TRAILING_BITS) {
+        break;
+      }
+
+      // Parse payload type
+      while (bytes[i] === 0xFF) {
+        payloadType += 255;
+        i++;
+      }
+      payloadType += bytes[i++];
+
+      // Parse payload size
+      while (bytes[i] === 0xFF) {
+        payloadSize += 255;
+        i++;
+      }
+      payloadSize += bytes[i++];
+
+      // this sei_message is a 608/708 caption so save it and break
+      // there can only ever be one caption message in a frame's sei
+      if (!result.payload && payloadType === USER_DATA_REGISTERED_ITU_T_T35) {
+        result.payloadType = payloadType;
+        result.payloadSize = payloadSize;
+        result.payload = bytes.subarray(i, i + payloadSize);
+        break;
+      }
+
+      // skip the payload and parse the next message
+      i += payloadSize;
+      payloadType = 0;
+      payloadSize = 0;
     }
-    result.payloadType = USER_DATA_REGISTERED_ITU_T_T35;
-
-    // parse the payload size
-    for (i = 2; i < bytes.length && bytes[i] === 0xff; i++) {
-      result.payloadSize += 255;
-    }
-    result.payloadSize <<= 8;
-    result.payloadSize |= bytes[i];
-    i++;
-
-    result.payload = bytes.subarray(i, i + result.payloadSize);
 
     return result;
   };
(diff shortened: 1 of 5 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -20,38 +20,62 @@
...
 
-    // parse the payload size
-    for (i = 2; i < bytes.length && bytes[i] === 0xff; i++) {
-      result.payloadSize += 255;
-    }
-    result.payloadSize <<= 8;
-    result.payloadSize |= bytes[i];
-    i++;
+      // Parse payload type
+      while (bytes[i] === 0xFF) {
```
dmlap: This could loop forever on malformed SEI messages.

lib/m2ts/caption-stream.js
```diff
@@ -20,38 +20,62 @@
...
+        payloadType += 255;
+        i++;
+      }
+      payloadType += bytes[i++];
 
-    result.payload = bytes.subarray(i, i + result.payloadSize);
+      // Parse payload size
+      while (bytes[i] === 0xFF) {
```
dmlap: Another potential infinite loop.

dmlap: ... or not. Nevermind these comments.

lib/m2ts/caption-stream.js
```diff
@@ -144,22 +175,35 @@
...
 
-    // send the data to the appropriate field
-    for (i = 0; i < captionPackets.length; i++) {
-      if (captionPackets[i].type === 0) {
-        this.field1_.push(captionPackets[i]);
-      }
+  CaptionStream.prototype.flush = function () {
+    // make sure we actually parsed captions before proceeding
+    if (!this.captionPackets_.length) {
+      this.field1_.flush();
```
dmlap: Would there be anything in field1 to flush if there weren't any caption packets received?

lib/m2ts/caption-stream.js
```diff
@@ -186,8 +229,8 @@
       ROLL_UP_2_ROWS             = 0x1425,
       ROLL_UP_3_ROWS             = 0x1426,
       ROLL_UP_4_ROWS             = 0x1427,
+      RESUME_DIRECT_CAPTIONING   = 0x1429,
```
dmlap: Currently unused, right?

lib/m2ts/caption-stream.js
```diff
@@ -275,17 +331,25 @@
...
-
       default:
-        charCode = data >>> 8;
+        char0 = data >>> 8;
+        char1 = data & 0xff;
+
+        // Look for a Channel 1 Preamble Address Code
+        if (char0 >= 0x10 && char0 <= 0x17 &&
+            char1 >= 0x40 && char1 <= 0x7F &&
+            (char0 !== 0x10 || char1 < 0x60)) {
```
dmlap: Nice and readable!

test/caption-stream-test.js
```diff
@@ -9,17 +9,38 @@
     }
   });
 
-  QUnit.skip('parses SEIs larger than 255 bytes', function() {
-    var captions = [];
-    captionStream.on('data', function(caption) {
-      captions.push(caption);
-    });
+  test('parses SEIs larger than 255 bytes', function() {
```
dmlap: Should this test description be updated? Looks like we're now also testing multi-message SEI packets.

test/caption-stream-test.js
```diff
@@ -232,43 +252,48 @@
...
+      // RCL, resume caption loading
+      { ccData: 0x1420 },
+      // '01'
+      { ccData: characters('01') },
+      // backspace
+      { ccData: 0x1421 },
+      { ccData: characters('23') },
+      // EOC, End of Caption
+      { pts: 1 * 1000, ccData: 0x142f },
+      // RCL, resume caption loading
```
dmlap: Might be worth a comment on why this control code is necessary between the two EOCs.

test/caption-stream-test.js
```diff
@@ -525,44 +571,107 @@
...
+      // backspace
+      { ccData: 0x1421 },
+      // CR, carriage return
+      { pts: 1 * 1000, ccData: 0x142d }
+    ].forEach(cea608Stream.push, cea608Stream);
 
     equal(captions.length, 0, 'no caption emitted');
   });
 
-  test('backspaces stop at the beginning of the line', function() {
```
dmlap: Looks like this test sort of got dropped.

test/caption-stream-test.js
```diff
@@ -525,44 +571,107 @@
...
+      {
+        pts: 0 * 1000,
+        ccData: characters('01')
+      },
+      // backspace
+      { ccData: 0x1421 },
+      // backspace
+      { ccData: 0x1421 },
+      // backspace
+      { ccData: 0x1421 },
```
dmlap: The duplicate control code filtering here would make this only two backspaces, right?

lib/m2ts/caption-stream.js
imbcmdth: This is just for completeness - flush on field1_ triggers `done` which triggers `done` on CaptionStream. I think streams should always propagate `done` events to any future stream even if they haven't emitted any data.

dmlap:
Should we have a test somewhere that verifies that SEI messages are properly escaped?

lib/m2ts/caption-stream.js
imbcmdth: Yes, I thought about removing it but I get the sneaking suspicion that we will be using it someday pretty soon.

lib/m2ts/caption-stream.js
imbcmdth: Heh.

lib/m2ts/caption-stream.js
imbcmdth: BUT THE CLOCK CYCLES!

test/caption-stream-test.js
imbcmdth: Just moved.

imbcmdth:
We already have a test for the `discardEmulationPreventionBytes` function - `properly parses seq_parameter_set_rbsp nal units` in transmuxer-test.js. While we don't have a function that tests `discardEmulationPreventionBytes` directly it *is* tested.

lib/m2ts/caption-stream.js
dmlap: Ok.

PR #43, opened 2015-12-09
BaseMediaDecodeTime & Audio Metadata fixes

imbcmdth:
1. Change how we calculate the baseMediaDecodeTime to correctly place segments without introducing a gap that start with a non-zero compositionTimeOffset for the first frame
1. Allow audio track metadata to change as we parse segments
 * Previously the emitted fragment would have metadata copied from the first segment we ever received and the meta data would never change even as the incoming segments changed audio parameters

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -73,6 +73,7 @@ AudioSegmentStream = function(track) {
         // If this is an allowed frame, keep it and record it's Dts
         if (currentFrame.dts >= earliestAllowedDts) {
           track.minSegmentDts = Math.min(track.minSegmentDts, currentFrame.dts);
+          track.minSegmentPts = track.minSegmentDts;
           return true;
         }
         // Otherwise, discard it
@@ -112,6 +113,7 @@ AudioSegmentStream = function(track) {
     boxes.set(mdat, moof.byteLength);
 
     clearDtsInfo(track);
+
     this.trigger('data', {track: track, boxes: boxes});
     this.trigger('done');
   };
@@ -131,6 +133,7 @@ VideoSegmentStream = function(track) {
     nalUnitsLength = 0,
     config,
     pps;
+
   VideoSegmentStream.prototype.init.call(this);
 
   delete track.minPTS;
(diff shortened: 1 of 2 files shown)


lib/mp4/transmuxer.js
```diff
@@ -73,6 +73,7 @@ AudioSegmentStream = function(track) {
         // If this is an allowed frame, keep it and record it's Dts
         if (currentFrame.dts >= earliestAllowedDts) {
           track.minSegmentDts = Math.min(track.minSegmentDts, currentFrame.dts);
+          track.minSegmentPts = track.minSegmentDts;
```
dmlap: Might be worth commenting that this is safe because audio frames won't be out of order.

lib/mp4/transmuxer.js
```diff
@@ -299,6 +302,18 @@ collectDtsInfo = function (track, data) {
     if (track.timelineStartInfo.pts === undefined) {
       track.timelineStartInfo.pts = data.pts;
     }
+
+    if (track.minSegmentPts === undefined) {
+      track.minSegmentPts = data.pts;
```
dmlap: Doesn't need to be resolved in this PR but I think having a notion of a "segment" here is probably not good.

test/transmuxer-test.js
```diff
@@ -1495,6 +1495,47 @@ test('calculates baseMediaDecodeTime values relative to a customizable baseMedia
...
+  videoSegmentStream.flush();
+
+  boxes = muxjs.tools.inspectMp4(segment);
+  tfdt = boxes[0].boxes[1].boxes[1];
+
+  // The timelineStartInfo's bMDT is 100 and that corresponds to a dts/pts of 10
+  // The first frame has a dts 50 so the bMDT is calculated as: (50 - 10) + 100 = 140
+  // The first frame has a compositionTimeOffset of: 60 - 50 = 10
+  // The final track's bMDT is therefore: 140 - 10 = 130
+  equal(tfdt.baseMediaDecodeTime, 130, 'calculated baseMediaDecodeTime');
```
dmlap: Great comment.

dmlap:
LGTM

PR #44, opened 2015-12-17
Turn Mux.js into a node.js module, and get it to work with browserify

BrandonOCasey:
* Merged old changes from @imbcmdth 
* Updated karma to use browserify when testing (so we don't have to build, then test)
* ignore build artifacts in git
* update all files to use require, rather than modifying globals


Gruntfile.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+module.exports = function (grunt) {
```
dmlap: @misteroneill has been pushing to standardize the rest of the video.js projects around pure npm-scripts for building. While mux.js isn't a traditional video.js plugin, I think it would be good to follow the same conventions unless we have a good reason to deviate.

debug/legacy.html
```diff
@@ -117,11 +107,6 @@ <h1 class="title">Transmux Analyzer</h1>
         vjsTags = document.querySelector('.vjs-tags'),
         workingTags = document.querySelector('.working-tags'),
         getOrderedTags,
-        tagTypes = {
-          0x08: 'audio',
-          0x09: 'video',
-          0x12: 'metadata'
-        },
```
dmlap: Good eye.

Gruntfile.js
imbcmdth: The build process is a left over from my changes. We may want to remove the build step altogether since it doesn't matter anymore. Whatever includes mux.js should have to worry about building it.

lib/flv/transmuxer.js
```diff
@@ -442,11 +447,9 @@ Transmuxer = function(options) {
     .pipe(this.metadataStream)
     .pipe(coalesceStream);
   // if CEA-708 parsing is available, hook up a caption stream
-  if (muxjs.mp2t.CaptionStream) {
-    captionStream = new muxjs.mp2t.CaptionStream();
-    h264Stream.pipe(captionStream)
-      .pipe(coalesceStream);
-  }
+  captionStream = new m2ts.CaptionStream();
```
dmlap: This looks like it makes caption parsing a requirement instead of an optional extension like it was previously.

lib/mp4/transmuxer.js
```diff
@@ -554,16 +556,17 @@ Transmuxer = function(options) {
...
-  parseStream = new muxjs.mp2t.TransportParseStream();
-  elementaryStream = new muxjs.mp2t.ElementaryStream();
-  aacStream = new muxjs.codecs.AacStream();
-  h264Stream = new muxjs.codecs.H264Stream();
+  packetStream = new m2ts.TransportPacketStream();
+  parseStream = new m2ts.TransportParseStream();
+  elementaryStream = new m2ts.ElementaryStream();
+  aacStream = new AacStream();
+  h264Stream = new H264Stream();
+  captionStream = new m2ts.CaptionStream();
```
dmlap: Also looks like captions are becoming a hard requirement here.

lib/tools/flv-inspector.js
```diff
@@ -1,13 +1,18 @@
...
+  },
   hex = function (val) {
     return '0x' + ('00' + val.toString(16)).slice(-2).toUpperCase();
   },
   hexStringList = function (data) {
-    var arr = [];
-
-    while(data.byteLength >0) {
+    var arr = [], i;
+    /* jshint -W086 */
```
dmlap: The alternative to this would be annotating each of the cases with something like `/* falls through */`, right? That's a bit more explicit but not a big deal.

lib/tools/mp4-inspector.js
```diff
@@ -216,7 +214,7 @@ var
       }
       // decode UTF-8 to javascript's internal representation
       // see http://ecmanaut.blogspot.com/2006/07/encoding-decoding-utf8-in-javascript.html
-      result.name = window.decodeURIComponent(window.escape(result.name));
+      result.name = decodeURIComponent(global.escape(result.name));
```
dmlap: Is escape() available in a node environment?

Gruntfile.js
misteroneill: Perhaps adopting some of the generator conventions could be done in a piecemeal way with this project in subsequent PRs. I was thinking the same about some of the bigger projects, which will not work with the generator's default setup (namely because they assume ES6 everything as does the linter).

lib/tools/mp4-inspector.js
gkatsev: yep

lib/tools/mp4-inspector.js
```diff
@@ -718,15 +728,24 @@ inspectMp4 = function(data) {
...
-    view = new DataView(data.buffer, data.byteOffset, data.byteLength),
+    view,
     size,
     type,
     end,
     box;
 
+  // DataView implementations differ in node & the browser
+  // so we check where this code is being run via a browserify variable
+  // process.browser
```
dmlap: This sounds like a bug in node. Is it something that will eventually be fixed?

package.json
```diff
@@ -17,16 +20,23 @@
...
     "hls",
     "mp4",
+    "flv",
     "aac",
     "h264"
   ],
   "author": "Brightcove",
   "license": "Apache-2.0",
   "devDependencies": {
+    "babelify": "^6.3.0",
```
dmlap: Are we using any es6 features? Do we need babelify?

package.json
```diff
@@ -17,16 +20,23 @@
     "player",
     "hls",
     "mp4",
+    "flv",
```
dmlap: I'm sure this is a hugely popular keyword on npm :)

test/index.html
```diff
@@ -10,7 +10,7 @@
   <div id="qunit-fixture"></div>
   <script src="../node_modules/qunitjs/qunit/qunit.js"></script>
 
-  <script src="../lib/utils/stream.js"></script>
+  <!--script src="../lib/utils/stream.js"></script>
```
dmlap: Please remove the scripts that are no longer being used.

test/index.html
```diff
@@ -21,19 +21,20 @@
   <script src="../lib/m2ts/metadata-stream.js"></script>
   <script src="../lib/mp4/transmuxer.js"></script>
   <script src="../lib/flv/flv-tag.js"></script>
-  <script src="../lib/flv/transmuxer.js"></script>
+  <script src="../lib/flv/transmuxer.js"></script-->
+  <script src="../dist/mux.js"></script>
 
   <!-- test helpers -->
-  <script src="test-segment.js"></script>
+  <!--script src="test-segment.js"></script>
```
dmlap: Same here.

test/index.html
```diff
@@ -21,19 +21,20 @@
...
   <script src="sintel-captions.js"></script>
-  <script src="id3-generator.js"></script>
+  <script src="id3-generator.js"></script-->
 
   <!-- test cases -->
-  <script src="exp-golomb-test.js"></script>
+  <!--script src="exp-golomb-test.js"></script>
```
dmlap: Same here.

test/mp4-generator-test.js
```diff
@@ -21,21 +20,18 @@
...
 var
-  mp4 = muxjs.mp4,
-  inspectMp4 = muxjs.tools.inspectMp4,
+  mp4 = require('../lib/mp4'),
   validateMvhd, validateTrak, validateTkhd, validateMdia,
   validateMdhd, validateHdlr, validateMinf, validateDinf,
   validateStbl, validateStsd, validateMvex,
   validateVideoSample, validateAudioSample;
 
-module('MP4 Generator');
```
dmlap: Intentional removal?

test/sintel-captions.js
```diff
@@ -1,3 +1,4 @@
+// Seems to be too big to work with browserify so we stick it in window
```
dmlap: Funny.

gkatsev: It may be possible to still use this in browserify by telling it to not parse this file. The option is called `noparse` in latest browserify: https://github.com/substack/node-browserify#browserifyfiles--opts

lib/flv/transmuxer.js
imbcmdth: With require this check didn't make sense (either you require it or you don't). Also there is little compelling reason to *not* include caption parsing support.

test/mp4-generator-test.js
BrandonOCasey: nope I added it back in, thanks for the catch!

dmlap:
Squashed and merged via db6ac31d2e286dea85cb985a96eb8ec023c8f3c1

PR #48, opened 2015-12-22
Clean up npm scripts

forbesjo:
Make some minor modifications to the npm scripts. `browserify` doesn't automatically make the dist/ directory and build isn't a built-in task.


dmlap:
LGTM

PR #52, opened 2016-01-21
Removed console.assert statements from code

tenacex:
Allows IE11 to run mux.js without erroring.


imbcmdth:
Be consistency!

Maintaining consistency is one of the most important things when working as part of a team.

You added three lines:

> ...should error handle

> ...handle error

> ...should handle error gracefully

PR #53, opened 2016-01-22
Remove window references in parsing functions of metadata stream

gesinger:
While the functions decodeURIComponent and unescape exist in the global scope, window is not available in all scopes (for instance, in the scope of a web worker).


dmlap:
@gesinger can you rebase this?

gesinger:
@dmlap no problem. Rebased.

PR #54, opened 2016-01-22
making development easier through the use of a dev-server and watchify

BrandonOCasey:
moving tests around to be more standardized, for easier access
added npm scripts for running a dev server, fixed linting
fixing serve issues
remove watch:js and build:test as these should never be needed
fixing a type in the build script
added standalone to browserify so there is a global muxjs in the browser, added clean

squashed/rebased version of #51


imbcmdth:
LGTM

PR #55, opened 2016-01-22
Added several errors and error stream emission

BrandonOCasey:

PR #56, opened 2016-01-25
added several errors and error stream emission

brandonocasey:
fixed some style issues, changed an else to an else if

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -291,6 +291,7 @@ TransportParseStream = function() {
 };
 TransportParseStream.prototype = new Stream();
 
+
 /**
  * Reconsistutes program elementary stream (PES) packets from parsed
  * transport stream packets. That is, if you pipe an
@@ -315,6 +316,84 @@ ElementaryStream = function() {
     timedMetadata = {
       data: [],
       size: 0
+    },
+    handlePesData = function(data) {
+      var stream, streamType;
+      switch (data.streamType) {
+      case StreamTypes.H264_STREAM_TYPE:
+        stream = video;
+        streamType = 'video';
+        break;
+      case StreamTypes.ADTS_STREAM_TYPE:
+        stream = audio;
+        streamType = 'audio';
+        break;
+      case StreamTypes.METADATA_STREAM_TYPE:
+        stream = timedMetadata;
+        streamType = 'timed-metadata';
+        break;
+      case undefined:
+        self.error('Found an undefined pes stream type');
+        return;
+      default:
+        self.error('Found an unknown pes stream type ' + data.streamType.toString(16));
+        return;
+      }
+
+      // if a new packet is starting, we can flush the completed
+      // packet
+      if (data.payloadUnitStartIndicator) {
+        flushStream(stream, streamType);
+      }
+
+      // buffer this fragment until we are sure we've received the
+      // complete payload
+      stream.data.push(data);
+      stream.size += data.data.byteLength;
+    },
+    handlePmtData = function(data) {
+      var event = {
+          type: 'metadata',
+          tracks: []
+        },
+        programMapTable = data.programMapTable,
+        k,
+        track;
+
+      // translate streams to tracks
+      for (k in programMapTable) {
+        if (programMapTable.hasOwnProperty(k)) {
+          track = {
+            timelineStartInfo: {
+              baseMediaDecodeTime: 0
+            }
+          };
+          track.id = +k;
+          if (programMapTable[k] === StreamTypes.H264_STREAM_TYPE) {
+            track.codec = 'avc';
+            track.type = 'video';
+          } else if (programMapTable[k] === StreamTypes.ADTS_STREAM_TYPE) {
+            track.codec = 'adts';
+            track.type = 'audio';
+          } else {
+            self.error('got some unknown kind ElementaryStream stream type: 0x' + programMapTable[k].toString(16));
+            continue;
+          }
+          event.tracks.push(track);
+        }
+      }
+
+      self.triggerData_(event);
+    },
+    handlePatData = function(data) {
+      // we have to wait for the PMT to arrive as well before we
+      // have any meaningful metadata
+      return;
+    },
+    handlers = {
+      'pes': handlePesData,
+      'pat': handlePatData,
+      'pmt': handlePmtData
     },
     parsePes = function(payload, pes) {
       var ptsDtsFlags;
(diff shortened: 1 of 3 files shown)


lib/m2ts/m2ts.js
```diff
@@ -41,11 +43,18 @@ TransportPacketStream = function() {
...
   this.push = function(bytes) {
     var
-      i = 0,
       startIndex = 0,
       endIndex = MP2T_PACKET_LENGTH,
       everything;
 
+    // if bytes is not a valid stream
+    if (typeof bytes !== 'object' ||
+       typeof bytes.byteLength === 'undefined' ||
```
dmlap: We gotta find a better way to test for `undefined`...

lib/m2ts/m2ts.js
```diff
@@ -231,20 +267,29 @@ TransportParseStream = function() {
...
     result.type = 'pes';
     result.data = packet.subarray(offset);
-
-    this.trigger('data', result);
+    this.triggerData_(result);
   };
 
 };
 TransportParseStream.prototype = new Stream();
-TransportParseStream.STREAM_TYPES  = {
```
dmlap: Where did these go?

lib/m2ts/m2ts.js
```diff
@@ -343,84 +390,101 @@ ElementaryStream = function() {
...
-              track.type = 'video';
-            } else if (programMapTable[k] === m2tsStreamTypes.ADTS_STREAM_TYPE) {
-              track.codec = 'adts';
-              track.type = 'audio';
+      // buffer this fragment until we are sure we've received the
+      // complete payload
+      stream.data.push(data);
+      stream.size += data.data.byteLength;
+    } else if (data.type === 'pmt') {
+      var
```
dmlap: The rest of this file sticks to one `var` declaration at the top of functions.

lib/m2ts/m2ts.js
```diff
@@ -343,84 +390,101 @@ ElementaryStream = function() {
...
+            continue;
           }
+          event.tracks.push(track);
         }
-        self.trigger('data', event);
       }
-    })[data.type]();
+
+      self.triggerData_(event);
+    } else if (data.type !== 'pat' && data.type !== 'pmt') {
```
dmlap: Didn't we already check for these types above?

lib/m2ts/m2ts.js
```diff
@@ -33,6 +32,9 @@ var
 TransportPacketStream = function() {
   var
     buffer = new Uint8Array(MP2T_PACKET_LENGTH),
+    totalDataEmitted = 0,
+    sinceFlushDataEmitted = 0,
+    invalidPushCalls = 0,
```
dmlap: Just for my background, what are we going to do with these stats?

lib/utils/stream.js
```diff
@@ -113,4 +116,8 @@ Stream.prototype.flush = function() {
   this.trigger('done');
 };
 
+Stream.prototype.error = function(error) {
```
dmlap: This method seems a bit incongruous with our usage of Stream for `data` events.

lib/m2ts/m2ts.js
```diff
@@ -343,84 +390,101 @@ ElementaryStream = function() {
...
-        default:
-          // ignore unknown stream types
-          return;
-        }
+    if (typeof data !== 'object' ||
+       !data.type) {
+      this.error('Data with no type passed to ElementaryStream');
+      return;
+    }
+    if (data.type === 'pat') {
```
dmlap: I'm not against re-organizing out of the object-property-dispatch here but if we're going to do that, I'd want to split these handlers out into helper functions as well. I think a big chain of if-elses is less readable than switching on well-defined tokens, though.

lib/m2ts/m2ts.js
BrandonOCasey: it was not actually used anywhere. we have all the stream types defined in a separate file called stream-types

lib/m2ts/m2ts.js
BrandonOCasey: https://github.com/videojs/mux.js/pull/55#discussion-diff-50763429

lib/m2ts/m2ts.js
BrandonOCasey: yeah that does make more sense, I will work on doing that

lib/m2ts/m2ts.js
```diff
@@ -269,6 +317,84 @@ ElementaryStream = function() {
       data: [],
       size: 0
     },
+    handlePesData = function(data) {
```
dmlap: You could make these private-ish methods and then skip the need for `self` (e.g. `handlePesData_` or just `pes_`).

lib/m2ts/m2ts.js
BrandonOCasey: as of right now there is not plan to do anything with the stats outside of checking for errors

lib/utils/stream.js
BrandonOCasey: I think nodejs streams can trigger an error of sorts as well https://nodejs.org/api/stream.html#stream_event_error

dmlap: Yes, they can. They do so via `stream.trigger(errorEvent)`, not through a custom method.

lib/m2ts/m2ts.js
dmlap: I think we should remove these until we have a need for them. Also, that means we do not need the layer of indirection between `stream.trigger(data)` so we can go back to using trigger() directly.

BrandonOCasey: I think that we do need one of these stats (sinceFlushDataEmitted) so that we can make sure that we emitted data before flush was called. I am fine getting rid of the others. This would mean keeping triggerData_ or overloading trigger, checking if the event is data and incrementing sinceFlushDataEmitted, then calling the prototype function with the same arguments. Unless we don't think that checking to see if we actually emitted data before a flush is an issue.

lib/utils/stream.js
BrandonOCasey: makes sense I will remove the error function from stream and change it for trigger('error', ...);

gkatsev:
Someone should take a look at whether the conflicts can be resolved in a reasonable amount of time (<30 min) otherwise, we should close this and perhaps schedule time to re-do it.

brandonocasey:
After looking this over, I don't think we want to use this. Most of the error cases in this pull request should probably be handled, rather than having an error on them. #224 is one example

PR #57, opened 2016-01-26
GOP-FUSION

imbcmdth:
The first iteration of holding onto the last-segment's-GOP and attempting to prepend that GOP to the next segment.

The `flush` functions are much nicer - far fewer variables; limited global mutations; fit on one screen! The bulk of the work has been moved to functions that perform limited parts of what the giant `while` loop used to do. The functions are written so that they do not mutate their parameters for sanity reasons.

While this might be slightly less performant, it is still O(n) and is much easier to reason about the steps that are happening on a flush.

The last GOP in a segment is held onto and prepended onto the next segment providing the next segment is greater than the GOP in time AND the next segment is within 10 seconds of the GOP.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -202,11 +202,8 @@ VideoSegmentStream = function(track) {
       // and the lastGop is within 10 seconds (90k * 10) of the first
       // nal unit
       if (ptsDistance > 0 && ptsDistance <= 900000) {
-        console.log('nals in last gop:', lastGop.length);
         lastGop.forEach(collectDtsInfo.bind(null, track));
         nalUnits = lastGop;
-      } else {
-        console.log('distance too large:', ptsDistance);
       }
 
       // 'delete' the lastGop
@@ -441,9 +438,9 @@ VideoSegmentStream = function(track) {
       sample.compositionTimeOffset = firstNal.pts - firstNal.dts;
 
       if (nextFrame) {
-        sample.duration = nextFrame[0].pts - firstNal.pts;
+        sample.duration = nextFrame[0].dts - firstNal.dts;
       } else {
-        sample.duration = lastNal.pts - firstNal.pts;
+        sample.duration = lastNal.dts - firstNal.dts;
 
         if (sample.duration === 0 &&
             samples.length > 0) {
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -345,12 +345,14 @@ VideoSegmentStream = function(track) {
     var
       i,
       currentFrame,
-      initialPts = frames[0][0].pts;
+      initialPts = frames[0][0].pts,
+      initialDts = frames[0][0].dts;
 
     for (i = 0; i < frames.length; i++) {
       currentFrame = frames[i];
       if (this.frameIsIFrame_(currentFrame)) {
         currentFrame[0].pts = initialPts;
+        currentFrame[0].dts = initialDts;
         break;
       }
     }
@@ -363,7 +365,7 @@ VideoSegmentStream = function(track) {
     var
       i,
       currentNal,
-      currentFrame,
+      currentFrame = [],
       frames = [];
 
     for (i = 0; i < nalUnits.length; i++) {
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -215,57 +215,6 @@ VideoSegmentStream = function(track) {
 
   this.gopCache = [];
 
-  this.getGopForFusion_ = function (nalUnit) {
-    var
-      halfSecond = 45000, // Half-a-second in a 90khz clock
-      allowableOverlap = 9009, // About 3 frames
-      nearestDistance = Infinity,
-      dtsDistance,
-      nearestGopObj,
-      currentGop,
-      currentGopObj,
-      i;
-
-    // Search for the GOP nearest to the beginning of this nal unit
-    for (i = 0; i < this.gopCache.length; i++) {
-      currentGopObj = this.gopCache[i];
-      currentGop = currentGopObj.gop;
-
-      // Reject Gops with different SPS or PPS
-      if (!(track.pps && compareArrays(track.pps[0], currentGopObj.pps[0])) ||
-          !(track.sps && compareArrays(track.sps[0], currentGopObj.sps[0]))) {
-        continue;
-      }
-
-      // Reject Gops that would require a negative baseMediaDecodeTime
-      if (currentGop.dts < track.timelineStartInfo.dts) {
-        continue;
-      }
-
-      dtsDistance = (nalUnit.dts - currentGop.dts) - currentGop.duration;
-
-      // Only consider GOPS that start before the nal unit and end within
-      // a half-second of the nal unit
-      if (dtsDistance >= -allowableOverlap &&
-          dtsDistance <= halfSecond) {
-
-        // Always use the closest GOP we found if there is more than
-        // one candidate
-        if (!nearestGopObj ||
-            nearestDistance > dtsDistance) {
-          nearestGopObj = currentGopObj;
-          nearestDistance = dtsDistance;
-        }
-      }
-    }
-
-    // If we found a GOP, use it to start the nalUnits array and update timeline info
-    if (nearestGopObj) {
-      return nearestGopObj.gop;
-    }
-    return false;
-  };
-
   this.push = function(nalUnit) {
     collectDtsInfo(track, nalUnit);
 
@@ -323,23 +272,39 @@ VideoSegmentStream = function(track) {
     frames = this.groupNalsIntoFrames_(nalUnits);
     gops = this.groupFramesIntoGops_(frames);
 
+    // If the first frame of this fragment is not an i-frame we have
+    // a problem since MSE (on Chrome) requires a leading i-frame.
+    //
+    // We have two approaches to repairing this situation:
+    // 1) GOP-FUSION:
+    //    This is where we keep track of the GOPS (group-of-pictures)
+    //    from previous fragments and attempt to find one that we can
+    //    prepend to the current fragment in order to create a valid
+    //    fragment.
+    // 2) I-FRAME-PULLING:
+    //    Here we search for the first i-frame in the fragment and
+    //    throw away all the frames between the start of the fragment
+    //    and that i-frame. We then extend the duration and pull the
+    //    PTS of the i-frame forward so that it covers the time range
+    //    of the frames that were disposed of.
+    //
+    // #1 is far prefereable over #2 which can cause "studdering" but
+    // requires more things to be just right.
     if (!gops[0][0].isIDR) {
+      // Search for a gop for fusion from our gopCache
       gopForFusion = this.getGopForFusion_(nalUnits[0], track);
 
       if (gopForFusion) {
         gops.unshift(gopForFusion);
-        // Adjust Gops' meta data to account for the inclusion
-        // of the new prefixed gop
+        // Adjust Gops' metadata to account for the inclusion of the
+        // new gop at the beginning
         gops.byteLength += gopForFusion.byteLength;
         gops.nalCount += gopForFusion.nalCount;
         gops.pts = gopForFusion.pts;
         gops.dts = gopForFusion.dts;
         gops.duration += gopForFusion.duration;
       } else {
-        // If we didn't find a candidate gop fall back to the
-        // old behavior of extending the first key frame to cover
-        // the portion of time left by removing all frames prior
-        // to the first i-frame
+        // If we didn't find a candidate gop fall back to i-frame-pulling
         gops = this.filterLeadingNonIFrames_(gops);
       }
     }
@@ -398,6 +363,59 @@ VideoSegmentStream = function(track) {
     pps = undefined;
   };
 
+  // Search for a candidate Gop for gop-fusion from the gop cache and
+  // return it or return null if no good candidate was found
+  this.getGopForFusion_ = function (nalUnit) {
+    var
+      halfSecond = 45000, // Half-a-second in a 90khz clock
+      allowableOverlap = 9009, // About 3 frames
+      nearestDistance = Infinity,
+      dtsDistance,
+      nearestGopObj,
+      currentGop,
+      currentGopObj,
+      i;
+
+    // Search for the GOP nearest to the beginning of this nal unit
+    for (i = 0; i < this.gopCache.length; i++) {
+      currentGopObj = this.gopCache[i];
+      currentGop = currentGopObj.gop;
+
+      // Reject Gops with different SPS or PPS
+      if (!(track.pps && compareArrays(track.pps[0], currentGopObj.pps[0])) ||
+          !(track.sps && compareArrays(track.sps[0], currentGopObj.sps[0]))) {
+        continue;
+      }
+
+      // Reject Gops that would require a negative baseMediaDecodeTime
+      if (currentGop.dts < track.timelineStartInfo.dts) {
+        continue;
+      }
+
+      // The distance between the end of the gop and the start of the nalUnit
+      dtsDistance = (nalUnit.dts - currentGop.dts) - currentGop.duration;
+
+      // Only consider GOPS that start before the nal unit and end within
+      // a half-second of the nal unit
+      if (dtsDistance >= -allowableOverlap &&
+          dtsDistance <= halfSecond) {
+
+        // Always use the closest GOP we found if there is more than
+        // one candidate
+        if (!nearestGopObj ||
+            nearestDistance > dtsDistance) {
+          nearestGopObj = currentGopObj;
+          nearestDistance = dtsDistance;
+        }
+      }
+    }
+
+    if (nearestGopObj) {
+      return nearestGopObj.gop;
+    }
+    return null;
+  };
+
   // create the default sample
   // see ISO/IEC 14496-12:2012, section 8.6.4.3
   this.createDefaultSample_ = function() {
@@ -441,8 +459,10 @@ VideoSegmentStream = function(track) {
     return gops;
   };
 
-  // convert an array of nal units into an array of frames with each frame being
+  // Convert an array of nal units into an array of frames with each frame being
   // composed of the nal units that make up that frame
+  // Also keep track of cummulative data about the frame from the nal units such
+  // as the frame duration, starting pts, etc.
   this.groupNalsIntoFrames_ = function(nalUnits) {
     var
       i,
@@ -457,7 +477,10 @@ VideoSegmentStream = function(track) {
     for (i = 0; i < nalUnits.length; i++) {
       currentNal = nalUnits[i];
 
+      // Split on 'aud'-type nal units
       if (currentNal.nalUnitType === 'access_unit_delimiter_rbsp') {
+        // Since the very first nal unit is expected to be an AUD
+        // only push to the frames array when currentFrame is not empty
         if (currentFrame && currentFrame.length) {
           currentFrame.duration = currentNal.dts - currentFrame.dts;
           frames.push(currentFrame);
@@ -467,6 +490,7 @@ VideoSegmentStream = function(track) {
         currentFrame.pts = currentNal.pts;
         currentFrame.dts = currentNal.dts;
       } else {
+        // Specifically flag i-frames for ease of use later
         if (currentNal.nalUnitType === 'slice_layer_without_partitioning_rbsp_idr') {
           currentFrame.isIDR = true;
         }
@@ -489,8 +513,10 @@ VideoSegmentStream = function(track) {
     return frames;
   };
 
-  // convert an array of frames into an array of Gop with each Gop being
-  // composed of the frames that make up that Gop
+  // Convert an array of frames into an array of Gop with each Gop being composed
+  // of the frames that make up that Gop
+  // Also keep track of cummulative data about the Gop from the frames such as the
+  // Gop duration, starting pts, etc.
   this.groupFramesIntoGops_ = function(frames) {
     var
       i,
@@ -498,11 +524,14 @@ VideoSegmentStream = function(track) {
       currentGop = [],
       gops = [];
 
+    // We must pre-set some of the values on the Gop since we
+    // keep running totals of these values
     currentGop.byteLength = 0;
     currentGop.nalCount = 0;
     currentGop.duration = 0;
     currentGop.pts = frames[0].pts;
     currentGop.dts = frames[0].dts;
+
     // store some fragment-level information
     gops.byteLength = 0;
     gops.nalCount = 0;
@@ -514,6 +543,8 @@ VideoSegmentStream = function(track) {
       currentFrame = frames[i];
 
       if (currentFrame.isIDR) {
+        // Since the very first frame is expected to be an i-frame
+        // only push to the gops array when currentGop is not empty
         if (currentGop && currentGop.length) {
           gops.push(currentGop);
           gops.byteLength += currentGop.byteLength;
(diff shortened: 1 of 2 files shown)


lib/mp4/transmuxer.js
```diff
@@ -20,7 +20,25 @@ var H264Stream = require('../codecs/h264').H264Stream;
...
+var collectDtsInfo, clearDtsInfo, calculateTrackBaseMediaDecodeTime, sumByteLengths;
+
+/**
+ * Sum the `byteLength` of a specific property in an array of objects
+ */
+sumByteLengths = function(array, property) {
+  var
+    i,
+    currentObj,
+    sum = 0;
```
dmlap: Man, we need to start switching to `let` in the project, pronto.

lib/mp4/transmuxer.js
```diff
@@ -49,29 +66,61 @@ AudioSegmentStream = function(track) {
...
+
+    this.trigger('data', {track: track, boxes: boxes});
+    this.trigger('done');
+  };
+
+  // If the audio segment extends before the earliest allowed dts
+  // value, remove AAC frames until starts at or after the earliest
+  // allowed DTS
+  this.trimAacFramesByEarliestDts_ = function(aacFrames) {
     if (track.minSegmentDts < earliestAllowedDts) {
```
dmlap: I would consider making `earliestAllowedDts` an argument to this function to make the dependencies clearer now that it's headed towards a general purpose utility. Not worth holding up the PR for, though.

lib/mp4/transmuxer.js
```diff
@@ -79,47 +128,49 @@ AudioSegmentStream = function(track) {
           return true;
         }
         // Otherwise, discard it
-        aacFramesLength -= currentFrame.data.byteLength;
         return false;
       });
+    } else {
+      return aacFrames;
```
dmlap: This clause should be first since it's so darn short.

lib/mp4/transmuxer.js
```diff
@@ -180,89 +257,30 @@ VideoSegmentStream = function(track) {
...
-      duration += sample.duration;
-      track.samples.shift();
-    }
+    // concatenate the video data and construct the mdat
+    mdat = mp4.mdat(this.concatenateNalData_(frames));
 
-    nalUnitsLength = 0;
-    mdat = mp4.mdat(data);
+    // Save all the nals in the last GOP for later
+    lastGop = gops.pop().reduce(function(a, b) { return a.concat(b); }, []);
```
dmlap: Are you just making a copy of the array with this reduce()? Why not slice()?

lib/mp4/transmuxer.js
```diff
@@ -291,7 +309,190 @@ VideoSegmentStream = function(track) {
...
+  this.filterLeadingNonIFrames_ = function(frames) {
+    var
+      i,
+      currentFrame,
+      initialPts = frames[0][0].pts;
+
+    for (i = 0; i < frames.length; i++) {
+      currentFrame = frames[i];
+      if (this.frameIsIFrame_(currentFrame)) {
+        currentFrame[0].pts = initialPts;
```
dmlap: This line deserves a comment.

lib/mp4/transmuxer.js
```diff
@@ -291,7 +309,190 @@ VideoSegmentStream = function(track) {
...
+      currentNal,
+      currentFrame,
+      frames = [];
+
+    for (i = 0; i < nalUnits.length; i++) {
+      currentNal = nalUnits[i];
+
+      if (currentNal.nalUnitType === 'access_unit_delimiter_rbsp') {
+        if (currentFrame && currentFrame.length) {
+          frames.push(currentFrame);
```
dmlap: If we made each element of `frames` an object instead, we could note whether it was a keyframe or not here without significant additional complexity.

lib/mp4/transmuxer.js
imbcmdth: It's `flatten` you luddite!

lib/mp4/transmuxer.js
```diff
@@ -140,38 +191,64 @@ VideoSegmentStream = function(track) {
...
+    var ptsDistance;
+
+    if (lastGop && lastGop.length) {
+      ptsDistance = nalUnit.pts - lastGop[0].pts;
+
+      // We only consider the lastGop valid for the current segment
+      // if the lastGop is earlier in time than the first nal unit
+      // and the lastGop is within 10 seconds (90k * 10) of the first
+      // nal unit
+      if (ptsDistance > 0 && ptsDistance <= 900000) {
```
dmlap: This check seems like an opportunity for improvement to me.

lib/mp4/transmuxer.js
```diff
@@ -140,38 +191,64 @@ VideoSegmentStream = function(track) {
 
   delete track.minPTS;
 
-  this.push = function(data) {
-    collectDtsInfo(track, data);
+  this.maybeUseLastGop_ = function (nalUnit) {
+    var ptsDistance;
+
+    if (lastGop && lastGop.length) {
```
dmlap: In discussion: what about checking `nalUnits.length === 0` instead?

dmlap:
One other thing to think about: what happens with `lastGop` if we're switching renditions to a encoding with a different SPS or PPS?

lib/mp4/transmuxer.js
```diff
@@ -119,7 +134,66 @@ AudioSegmentStream = function(track) {
...
+    return samples;
+  };
+
+  // generate the track's sample table from an array of frames
+  this.concatenateFrameData_ = function(frames) {
+    var
+      i,
+      currentFrame,
+      dataOffset = 0,
+      data = new Uint8Array(sumByteLengths(frames, 'data'));
```
dmlap: I'm still slightly concerned that we're doing byte-wise iteration over and over again. Is there any way we could test the performance impact of this?

lib/mp4/transmuxer.js
```diff
@@ -140,38 +214,102 @@ VideoSegmentStream = function(track) {
...
+          !(track.sps && compareArrays(track.sps[0], currentGopObj.sps[0]))) {
+        continue;
+      }
+
+      dtsDistance = nalUnit.dts - currentGopNals[currentGopNals.length - 1].dts;
+
+      // Only consider GOPS that start before the nal unit and end within
+      // a half-second of the nal unit
+      if (nalUnit.dts >= currentGopNals[0].dts &&
+        dtsDistance <= 45000) {
```
dmlap: 45000 should be a constant somewhere.

lib/mp4/transmuxer.js
```diff
@@ -140,38 +214,102 @@ VideoSegmentStream = function(track) {
...
+      moof,
+      mdat,
+      boxes;
+
+    // Search for a candidate GOP for FUSION (prepending the gop to this segment)
+    gopForFusion = this.getGopForFusion_(nalUnits[0]);
+
+    // Prepend it if we found it
+    nalUnits = gopForFusion.concat(nalUnits);
+    // Prepend it if we found it
```
dmlap: Repeated comment from above

lib/mp4/transmuxer.js
```diff
@@ -291,7 +381,197 @@ VideoSegmentStream = function(track) {
...
+
+  // generate the track's raw mdat data from an array of frames
+  this.concatenateNalData_ = function (frames) {
+    var
+      i, j,
+      currentFrame,
+      currentNal,
+      dataOffset = 0,
+      nalsByteLength = frames.reduce(function(v, frame) {return v + sumByteLengths(frame, 'data'); }, 0),
+      numberOfNals = frames.reduce(function(v, frame) { return v + frame.length; }, 0),
```
dmlap: These two reduces() seem like more points in favor of building up and storing information about frames, rather than recalculate it over and over.

test/transmuxer.test.js
```diff
@@ -1265,49 +1310,346 @@ QUnit.test('infers sample durations from DTS values', function() {
 
   boxes = mp4.tools.inspect(segment);
   samples = boxes[0].boxes[1].boxes[2].samples;
-  QUnit.equal(samples.length, 3, 'generated three samples');
-  QUnit.equal(samples[0].duration, 1, 'set the first sample duration');
-  QUnit.equal(samples[1].duration, 2, 'set the second sample duration');
-  QUnit.equal(samples[2].duration, 2, 'inferred the final sample duration');
+  QUnit.equal(samples.length, 2, 'generated two samples, filters out pre-IDR');
+  QUnit.equal(samples[0].duration, 3, 'set the first sample duration');
+  QUnit.equal(samples[1].duration, 3, 'set the second sample duration');
```
dmlap: This duration is three because we re-use the last known duration for the final sample?

test/transmuxer.test.js
```diff
@@ -1265,49 +1310,346 @@ QUnit.test('infers sample durations from DTS values', function() {
...
 });
 
-QUnit.test('filters pre-IDR samples and caluculate duration correctly', function() {
+QUnit.test('holds onto the last GOP and prepends the subsequent push operation with that GOP', function() {
   var segment, boxes, samples;
+
+  videoSegmentStream.push({
```
dmlap: I had been avoiding this but creating all the test NALs for these tests below seem to distract from the point of the test. What do you think about a helper function of some sort?

lib/mp4/transmuxer.js
```diff
@@ -140,38 +214,102 @@ VideoSegmentStream = function(track) {
 
   delete track.minPTS;
 
-  this.push = function(data) {
-    collectDtsInfo(track, data);
+  this.getGopForFusion_ = function (nalUnit) {
```
dmlap: This function definitely needs a comment. I think it is worth thinking about a more descriptive name. "GOP fusion" sounds really cool but I would have a hard time guessing what that means. What about findClosestGop()?

lib/mp4/transmuxer.js
imbcmdth: We aren't doing byte-wise iteration at all. Instead simply iterating over each *frame* to sum the lengths. It's exactly the same work as before just in it's own function. Also this is audio which is an order of magnitude less data (at least) than the video side.

test/transmuxer.test.js
imbcmdth: Yes.

lib/mp4/transmuxer.js
```diff
@@ -20,7 +20,48 @@ var H264Stream = require('../codecs/h264').H264Stream;
 var VideoSegmentStream, AudioSegmentStream, Transmuxer, CoalesceStream;
 
 // Helper functions
-var collectDtsInfo, clearDtsInfo, calculateTrackBaseMediaDecodeTime;
+var collectDtsInfo, clearDtsInfo, calculateTrackBaseMediaDecodeTime, compareArrays, sumByteLengths;
+
+/**
+ * Compare two arrays (event typed) for same-ness
+ */
+compareArrays = function(a, b) {
```
dmlap: This function name is a bit vague. What about arrayEquals()?

lib/mp4/transmuxer.js
```diff
@@ -20,7 +20,48 @@ var H264Stream = require('../codecs/h264').H264Stream;
...
+  var
+    i,
+    currentObj,
```
dmlap: Is `currentObj` used?

lib/mp4/transmuxer.js
```diff
@@ -20,7 +20,48 @@ var H264Stream = require('../codecs/h264').H264Stream;
...
+    sum = 0;
+
+  if (a.length !== b.length) {
+    return false;
+  }
+
+  // sum the byteLength's all each nal unit in the frame
```
dmlap: This comment seems out of place.

lib/mp4/transmuxer.js
```diff
@@ -20,7 +20,48 @@ var H264Stream = require('../codecs/h264').H264Stream;
...
+    if (a[i] !== b[i]) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
+/**
+ * Sum the `byteLength` of a specific property in an array of objects
```
dmlap: This could also be done as a map() then a sum. If you like sticking with this method, could you elaborate a bit more on this function level comment?

lib/mp4/transmuxer.js
```diff
@@ -119,7 +134,66 @@ AudioSegmentStream = function(track) {
     this.trigger('data', {track: track, boxes: boxes});
     this.trigger('done');
   };
+
+  // If the audio segment extends before the earliest allowed dts
+  // value, remove AAC frames until starts at or after the earliest
+  // allowed DTS
```
dmlap: I think we should explain why there is an earliest allowed DTS in the first place. We might have a comment somewhere else but it wouldn't hurt to repeat ourselves a bit in comment form.

lib/mp4/transmuxer.js
dmlap: Or even better, just a map().

lib/mp4/transmuxer.js
```diff
@@ -119,7 +134,66 @@ AudioSegmentStream = function(track) {
...
+
+  // generate the track's sample table from an array of frames
+  this.concatenateFrameData_ = function(frames) {
+    var
+      i,
+      currentFrame,
+      dataOffset = 0,
+      data = new Uint8Array(sumByteLengths(frames, 'data'));
+
+    for (i = 0; i < frames.length; i++) {
```
dmlap: Not necessary for this PR but I think we concatenate typed arrays in a bunch of places. Seems worth consolidating into a utility to me.

lib/mp4/transmuxer.js
```diff
@@ -179,90 +259,76 @@ VideoSegmentStream = function(track) {
...
+    //    prepend to the current fragment in order to create a valid
+    //    fragment.
+    // 2) I-FRAME-PULLING:
+    //    Here we search for the first i-frame in the fragment and
+    //    throw away all the frames between the start of the fragment
+    //    and that i-frame. We then extend the duration and pull the
+    //    PTS of the i-frame forward so that it covers the time range
+    //    of the frames that were disposed of.
+    //
+    // #1 is far prefereable over #2 which can cause "studdering" but
```
dmlap: Studdering or *stuttering*?

lib/mp4/transmuxer.js
```diff
@@ -179,90 +259,76 @@ VideoSegmentStream = function(track) {
...
+    // requires more things to be just right.
+    if (!gops[0][0].isIDR) {
```
dmlap: I believe this should be `isIdr` by our capitalization rules. I also think we need to be much more consistent with terminology. The FLV transmuxer has a boolean `keyFrame` on frame objects. You also refer to this as an "i-frame" in your comments above and don't mention IDRs anywhere. Let's choose one and stick with it.

lib/mp4/transmuxer.js
```diff
@@ -179,90 +259,76 @@ VideoSegmentStream = function(track) {
...
+        // Adjust Gops' metadata to account for the inclusion of the
+        // new gop at the beginning
+        gops.byteLength += gopForFusion.byteLength;
+        gops.nalCount += gopForFusion.nalCount;
+        gops.pts = gopForFusion.pts;
+        gops.dts = gopForFusion.dts;
+        gops.duration += gopForFusion.duration;
+      } else {
+        // If we didn't find a candidate gop fall back to i-frame-pulling
+        gops = this.filterLeadingNonIFrames_(gops);
```
dmlap: And here we call them iframes :)

lib/mp4/transmuxer.js
```diff
@@ -179,90 +259,76 @@ VideoSegmentStream = function(track) {
...
-      sample = track.samples[0];
-      if (sample.flags.dependsOn === 2) {
-        data = data.subarray(sample.dataOffset);
-        sample.duration += duration;
-        break;
-      }
-      duration += sample.duration;
-      track.samples.shift();
-    }
+    // Keep a maximum of 6 GOPs in the cache (about a minute worth)
```
dmlap: "about a minute worth" isn't really accurate with variable segment sizes.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
-    clearDtsInfo(track);
     this.trigger('data', {track: track, boxes: boxes});
 
+    this.resetStreamStage_();
+
+    // Continue with the flush process now
+    this.trigger('done');
+  };
+
+  this.resetStreamStage_ = function() {
```
dmlap: I don't understand what "resetStreamStage_()" means. What is a stream stage?

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+        continue;
+      }
+
+      // Reject Gops that would require a negative baseMediaDecodeTime
+      if (currentGop.dts < track.timelineStartInfo.dts) {
+        continue;
+      }
+
+      // The distance between the end of the gop and the start of the nalUnit
+      dtsDistance = (nalUnit.dts - currentGop.dts) - currentGop.duration;
```
dmlap: There might be something wrong here. Duration is calculated based on PTS. DTS + duration is a weird calculation.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+  };
 
-    // Continue with the flush process now
-    this.trigger('done');
+  // Search for a candidate Gop for gop-fusion from the gop cache and
+  // return it or return null if no good candidate was found
+  this.getGopForFusion_ = function (nalUnit) {
+    var
+      halfSecond = 45000, // Half-a-second in a 90khz clock
+      allowableOverlap = 9009, // About 3 frames
```
dmlap: Not true for different frame rates.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+
+    if (nearestGopObj) {
+      return nearestGopObj.gop;
+    }
+    return null;
+  };
+
+  // create the default sample
+  // see ISO/IEC 14496-12:2012, section 8.6.4.3
+  this.createDefaultSample_ = function() {
```
dmlap: This might be a little easier to understand as an object prototype/class that you can `new`.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+  this.filterLeadingNonIFrames_ = function(gops) {
+    var
+      h, i,
+      currentGop,
+      newGops;
+
+    // Only filter leading i-frames if we have more than one frame
+    // This is just a concession to make tests work. Since most tests
+    // don't bother creating leading iframes, they would usually get
+    // filtered ruining the test.
```
dmlap: We should fix this.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+        isLeading: 0,
+        dependsOn: 1,
+        isDependedOn: 0,
+        hasRedundancy: 0,
+        degradationPriority: 0
+      }
+    };
+  };
+
+  this.filterLeadingNonIFrames_ = function(gops) {
```
dmlap: I think this function name is a bit misleading now that it operates on gops. What about something like extendLeadingGop()?

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+    currentFrame.byteLength = 0;
+
+    for (i = 0; i < nalUnits.length; i++) {
+      currentNal = nalUnits[i];
+
+      // Split on 'aud'-type nal units
+      if (currentNal.nalUnitType === 'access_unit_delimiter_rbsp') {
+        // Since the very first nal unit is expected to be an AUD
+        // only push to the frames array when currentFrame is not empty
+        if (currentFrame && currentFrame.length) {
```
dmlap: It doesn't appear that currentFrame will ever be null or undefined.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+    gops.pts = frames[0].pts;
+    gops.dts = frames[0].dts;
+
+    for (i = 0; i < frames.length; i++) {
+      currentFrame = frames[i];
+
+      if (currentFrame.isIDR) {
+        // Since the very first frame is expected to be an i-frame
+        // only push to the gops array when currentGop is not empty
+        if (currentGop && currentGop.length) {
```
dmlap: Also doesn't appear `currentGop` could be falsey.

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+
+    // We must pre-set some of the values on the Gop since we
+    // keep running totals of these values
+    currentGop.byteLength = 0;
+    currentGop.nalCount = 0;
+    currentGop.duration = 0;
+    currentGop.pts = frames[0].pts;
+    currentGop.dts = frames[0].dts;
+
+    // store some fragment-level information
```
dmlap: What's a fragment?

lib/mp4/transmuxer.js
```diff
@@ -140,38 +213,45 @@ VideoSegmentStream = function(track) {
 
   delete track.minPTS;
 
-  this.push = function(data) {
-    collectDtsInfo(track, data);
+  this.gopCache = [];
```
dmlap: This should be a private property, right?

lib/mp4/transmuxer.js
```diff
@@ -274,24 +340,318 @@ VideoSegmentStream = function(track) {
...
+
+      // Only consider GOPS that start before the nal unit and end within
+      // a half-second of the nal unit
+      if (dtsDistance >= -allowableOverlap &&
+          dtsDistance <= halfSecond) {
+
+        // Always use the closest GOP we found if there is more than
+        // one candidate
+        if (!nearestGopObj ||
+            nearestDistance > dtsDistance) {
```
dmlap: Does this actually happen?

lib/mp4/transmuxer.js
imbcmdth: It's used (well, cleared) by the Transmuxer function so I don't consider it private.

lib/mp4/transmuxer.js
imbcmdth: It is certainly possible with a *very* short segment. I wanted to account for the possibility and always select the best candidate.

lib/mp4/transmuxer.js
imbcmdth: `duration` is actually calculated from `dts`. AFAIK, it has always been: https://github.com/videojs/mux.js/blob/master/lib/mp4/transmuxer.js#L215

I actually had a bug early on where I used `pts` instead of `dts` for calculating the `duration` and the results were very, very wrong when given a ts with b-frames.

lib/mp4/transmuxer.js
dmlap: Do you foresee a need for an outside part to manipulate the cache other than clearing it?

lib/mp4/transmuxer.js
```diff
@@ -49,60 +108,35 @@ AudioSegmentStream = function(track) {
...
 
     // concatenate the audio data to constuct the mdat
-    data = new Uint8Array(aacFramesLength);
-    track.samples = [];
-    i = 0;
-    while (aacFrames.length) {
-      currentFrame = aacFrames[0];
-      sample = {
-        size: currentFrame.data.byteLength,
-        duration: 1024 // FIXME calculate for realz
```
dmlap: Kinda sad to see this delightful comment go...

lib/mp4/transmuxer.js
```diff
@@ -274,24 +360,303 @@ VideoSegmentStream = function(track) {
...
+    var
+      h, i,
+      currentGop,
+      newGops;
+
+    // Only filter leading keyframes if we have more than one frame
+    // This is just a concession to make tests work. Since most tests
+    // don't bother creating leading iframes, they would usually get
+    // filtered ruining the test.
+    if (!gops[0][0].keyFrame && gops.length > 1) {
```
dmlap: I think we should try to fix this before merging.

PR #58, opened 2016-01-28
Aac file parsing

tenacex:
This PR includes support for AAC file parsing. It works with the existing mp4 transmuxer and can successfully output an audio segment. As per request, it currently does not support flv. The debug page has also been extended to support AAC file input, which it can detect automatically.


debug/index.html
```diff
@@ -138,6 +138,7 @@ <h1 class="title">Transmux Analyzer</h1>
     var inputs = document.getElementById('inputs'),
         original = document.getElementById('original'),
         working = document.getElementById('working'),
+        original_aac = document.getElementById('original-aac'),
```
imbcmdth: Is this used anywhere?

debug/index.html
```diff
@@ -326,7 +330,6 @@ <h1 class="title">Transmux Analyzer</h1>
 
       reader.addEventListener('loadend', function() {
         var bytes = new Uint8Array(reader.result);
-
```
imbcmdth: You keep removing whitespace! The careful use of whitespace (in this case, after variable definitions) adds some readability to the code.

lib/AACStream/stream-types.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
imbcmdth: Why was this file copied to the AACStream directory?

lib/AACStream/metadata-stream.js
```diff
@@ -0,0 +1,229 @@
+/**
```
imbcmdth: Why was this file copied to the AACStream directory?

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
+/**
```
imbcmdth: Following the naming conventions in the rest of the project, this file should be `aac.js` and be located in a directory called `aac`. You could also just make this file be the index.js in `aac`, I would be ok with that and it'll reduce the file count by one.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+ * deliver mp4s to a SourceBuffer on platforms that support native
+ * Media Source Extensions.
+ */
+'use strict';
+var Stream = require('../utils/stream.js');
+
+// object types
+var AacStream;
+
+var
```
imbcmdth: Because of the way `var` is being used here, you are defining AacStream twice. I suspect one is just leftover.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Splits an incoming stream of binary data into ADTS and ID3 Frames.
+ */
+
+ AacStream = function() {
+  var
+    everything,
+    frameSize = 0,
```
imbcmdth: Why is `frameSize` defined here if it is only used in `this.push`?

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+              data: everything.subarray(0, frameSize)
+            };
+            this.trigger('data', chunk);
+            everything = everything.subarray(frameSize);
+            continue;
+          } else {
+            canParseMore = false;
+          }
+        } else if ((everything[0] & 0xff == 0xff) &&
+            (((everything[1] >> 4) & 0xf) == 0xf)) {
```
imbcmdth: While it doesn't matter here, you almost always want the zero-filling right shift `>>>` instead of the sign-extending right shift. This difference can bite you in subtle ways when dealing with large numbers and the zero-filling one is generally "safer". Also you can just mask the upper 4 bits and do: `(everything[1] & 0xF0) === 0xF0`.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+              type: 'timed-metadata',
+              data: everything.subarray(0, frameSize)
+            };
+            this.trigger('data', chunk);
+            everything = everything.subarray(frameSize);
+            continue;
+          } else {
+            canParseMore = false;
+          }
+        } else if ((everything[0] & 0xff == 0xff) &&
```
imbcmdth: You are using the unsafe type-coercing equality operator here. ALWAYS use `===` and forget you ever knew about `==` when doing JavaScript.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+    var d = frame.data;
+    var size = ((d[3] & 0x01)  << 30) | (d[4]  << 22) | (d[5] << 14) | (d[6] << 6) | d[7] >>> 2;
+    size *= 4;
+    size += d[7] & 0x03;
+    timeStamp = size;
+  };
+
+
+  this.push = function(bytes) {
+    var canParseMore = true;
```
imbcmdth: Is there no way to write this loop without using a boolean variable?

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+  var
+    everything,
+    frameSize = 0,
+    receivedTimeStamp = false,
+    timeStamp = 0;
+
+  AacStream.prototype.init.call(this);
+
+  this.setTimestamp = function (frame) {
+    var d = frame.data;
```
imbcmdth: You do this several times. Right now this project is using es5 conventions and our style guide proscribes that we only use **one** `var` statement at the top of the function always.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+    var size = ((d[3] & 0x01)  << 30) | (d[4]  << 22) | (d[5] << 14) | (d[6] << 6) | d[7] >>> 2;
+    size *= 4;
+    size += d[7] & 0x03;
+    timeStamp = size;
```
imbcmdth: I prefer it if functions didn't mutate outside state. This function should just return a value.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+
+  this.push = function(bytes) {
+    var canParseMore = true;
+
+    // If there are bytes remaining from the last segment, prepend them to the
+    // bytes that were pushed in
+    if (everything !== undefined && everything.length) {
+      var tempLength = everything.length;
```
imbcmdth: This `var` should be at the top of the function.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+      everything.set(bytes, tempLength);
+    } else {
+      everything = bytes;
+    }
+
+    while(canParseMore) {
+      if (everything.length >= 10) {
+        if ((everything[0] === 'I'.charCodeAt(0)) &&
+          (everything[1] === 'D'.charCodeAt(0)) &&
+          (everything[2] === '3'.charCodeAt(0))) {
```
imbcmdth: I think this code would be much cleaner if ID3 tag parsing and ADTS frame parsing were separate functions called based on what the first few bytes are.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+            canParseMore = false;
+          }
+        }
+      } else {
+        canParseMore = false;
+      }
+    }
+  };
+  this.flush = function () {
+    this.trigger('done');
```
imbcmdth: This trigger is performed implicitly if you don't define a `flush` function. Just delete this function.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+  };
+};
+
+AacStream.prototype = new Stream();
+
+var AACStream = {
+  AacStream: AacStream,
+  MetadataStream: require('./metadata-stream'),
```
imbcmdth: Why are requiring a different `metadata-stream.js` than the usual one?

lib/AACStream/metadata-stream.js
tenacex: This is a slightly altered metadata-stream since it looks for the timestamp ID3 tags and then sends them backward through the pipeline. I can however alter the file in the m2ts folder, but then I believe the file would be in an incorrect folder since it is being used by files outside the m2ts directory.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: Moved inside of push.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: This function is the function that runs when we receive a 'timestamp' event. How can it be affective if it doesn't alter the state outside of it?

lib/AACStream/aacs.js
tenacex: changed.

lib/AACStream/aacs.js
tenacex: I pulled out the framesize calculations which took up a lot of space and made it hard to read. I left in the actual construction of the packet/chunk to keep with your mentality of not having functions modify outside of their scope. Let me know if you want it modularized further.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: This is a slightly altered metadata-stream since it looks for the timestamp ID3 tags and then sends them backward through the pipeline. I can however alter the file in the m2ts folder, but then I believe the file would be in an incorrect folder since it is being used by files outside the m2ts directory.

debug/index.html
tenacex: Removed.

debug/index.html
tenacex: Re-added.

lib/AACStream/stream-types.js
tenacex: Removed from the next commit.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+      var tempLength = everything.length;
+      everything = new Uint8Array(bytes.byteLength + tempLength);
+      everything.set(everything.subarray(0, tempLength));
+      everything.set(bytes, tempLength);
+    } else {
+      everything = bytes;
+    }
+
+    while(true) {
+      if (everything.length >= 10) {
```
imbcmdth: Why not make the `while` condition equal to this `if`'s condition? Then you won't have the danger of infinite iteration looming over you! Also there should be a space between the keyword and the opening parens: `while (...`

lib/mp4/transmuxer.js
```diff
@@ -546,51 +547,103 @@ Transmuxer = function(options) {
...
-    .pipe(aacStream);
-
-  elementaryStream
-    .pipe(this.metadataStream)
-    .pipe(coalesceStream);
-
-  // Hook up CEA-608/708 caption stream
-  h264Stream.pipe(captionStream)
-    .pipe(coalesceStream);
+    //audioSegmentStream = new AudioSegmentStream();
```
imbcmdth: Leftover?

lib/mp4/transmuxer.js
```diff
@@ -546,51 +547,103 @@ Transmuxer = function(options) {
...
+          .pipe(audioSegmentStream)
+          .pipe(coalesceStream);
+
+
+      }
+    }
+  });
+  };
+
+    this.setupTSPipeline = function() {
```
imbcmdth: Indenting here looks wonky.

lib/mp4/transmuxer.js
```diff
@@ -546,51 +547,103 @@ Transmuxer = function(options) {
...
+    this.metadataStream.pipe(coalesceStream);
+
+    this.metadataStream.on('timestamp', aacStream.setTimestamp.bind(aacStream));
+    this.addAACListener();
+
+  };
+
+
+  this.addAACListener = function() {
+      aacStream.on('data', function(data) {
```
imbcmdth: Indenting looks wonky.

lib/mp4/transmuxer.js
```diff
@@ -546,51 +547,103 @@ Transmuxer = function(options) {
...
+      if (track && !audioSegmentStream) {
+        // hook up the audio segment stream to the first track with aac data
+        coalesceStream.numberOfTracks++;
+        audioSegmentStream = new AudioSegmentStream(track);
+
+        // Set up the final part of the audio pipeline
+        adtsStream
+          .pipe(audioSegmentStream)
+          .pipe(coalesceStream);
+
```
imbcmdth: There are extra empty newlines in this file... is this where all the ones that were missing in the other file snuck off to? :smile:

lib/mp4/transmuxer.js
```diff
@@ -638,13 +691,37 @@ Transmuxer = function(options) {
...
           .pipe(coalesceStream);
       }
     }
   });
 
+  };
+
+  Transmuxer.prototype.init.call(this);
+  options = options || {};
+  aacfile = options.aacfile;
```
imbcmdth: Does this need to be a variable? You can just do `option.aacfile === undefined` in the one place it is used. Also consider naming this something like `aacInput`.

lib/mp4/transmuxer.js
```diff
@@ -663,13 +740,23 @@ Transmuxer = function(options) {
 
   // feed incoming data to the front of the parsing pipeline
   this.push = function(data) {
-    packetStream.push(data);
+    if (aacfile === undefined) {
```
imbcmdth: You can get rid of this whole condition if you just saved whatever object was the head of the pipeline to a single variable. In other words, on push you should always push data into `pipelineRoot` or some such. Which object it actually is will be determined by which setup function was called.

lib/mp4/transmuxer.js
```diff
@@ -663,13 +740,23 @@ Transmuxer = function(options) {
...
+      aacStream.push(data);
+    }
+
   };
 
   // flush any buffered data
   this.flush = function() {
     // Start at the top of the pipeline and flush all pending work
-    packetStream.flush();
+    if (aacfile === undefined) {
```
imbcmdth: Same here.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+
+  this.parseID3 = function(everything) {
+    var returnSize = (everything[6] << 21) | (everything[7] << 14) |
+              (everything[8] << 7) | (everything[9]),
+      flags = everything[5],
+      footerpresent = (flags & 16) >> 4;
+    if (footerpresent) {
+            return returnSize + 20;
+    }
+    return returnSize + 10;t
```
imbcmdth: `return returnSize + 10;t` ??

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+  this.parseADTS = function(everything) {
+    var lowThree = (everything[5] & 0xE0) >> 5,
+      middle = everything[4] << 3,
+      highTwo = everything[3] & 0x3 << 11;
+      return (highTwo | middle) | lowThree;
+  };
+
+
+  this.push = function(bytes) {
+    var canParseMore = true,
```
imbcmdth: Is `canParseMore` used anymore?

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+        if ((everything[0] === 'I'.charCodeAt(0)) &&
+          (everything[1] === 'D'.charCodeAt(0)) &&
+          (everything[2] === '3'.charCodeAt(0))) {
+
+          //check framesize
+          frameSize = this.parseID3(everything);
+
+          //we have enough in the buffer to emit a full packet
+          if(everything.length >= frameSize) {
+            var chunk = {
```
imbcmdth: Style guide specifies that you can only have one `var` at the top of functions.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,123 @@
...
+          } else {
+            break;
+          }
+        } else if ((everything[0] & 0xff === 0xff) &&
+            ((everything[1] & 0xf0) === 0xf0)) {
+
+          frameSize = this.parseADTS(everything);
+
+          if(everything.length >= frameSize) {
+            var packet = {
```
imbcmdth: Style guide specifies that you can only have one `var` at the top of functions.

lib/AACStream/aacs.js
tenacex: Changed. Thanks for that tip.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed... I think?

lib/mp4/transmuxer.js
tenacex: It is used in 3 other places in this file. Don't you think it deserves variable status?

lib/AACStream/aacs.js
tenacex: Changed, snuck in there somehow.

lib/AACStream/aacs.js
tenacex: Removed.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+    while (everything.length >= 10) {
+      if ((everything[0] === 'I'.charCodeAt(0)) &&
+        (everything[1] === 'D'.charCodeAt(0)) &&
+        (everything[2] === '3'.charCodeAt(0))) {
+
+        //check framesize
+        frameSize = this.parseID3(everything);
+
+        //we have enough in the buffer to emit a full packet
+        if(everything.length >= frameSize) {
```
imbcmdth: Missing a space after the `if`.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+          continue;
+        } else {
+          break;
+        }
+      } else if ((everything[0] & 0xff === 0xff) &&
+          ((everything[1] & 0xf0) === 0xf0)) {
+
+        frameSize = this.parseADTS(everything);
+
+        if(everything.length >= frameSize) {
```
imbcmdth: Missing a space after the `if`.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+          var packet = {
```
imbcmdth: Style guide specifies that you can only have one var and it must appear at the top of functions.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+      if ((everything[0] === 'I'.charCodeAt(0)) &&
+        (everything[1] === 'D'.charCodeAt(0)) &&
+        (everything[2] === '3'.charCodeAt(0))) {
+
+        //check framesize
+        frameSize = this.parseID3(everything);
+
+        //we have enough in the buffer to emit a full packet
+        if(everything.length >= frameSize) {
+          var chunk = {
```
imbcmdth: Style guide specifies that you can only have one var and it must appear at the top of functions.

imbcmdth:
Missing tests!

lib/AACStream/aacs.js
tenacex: Fixed.

lib/AACStream/aacs.js
tenacex: Fixed.

lib/AACStream/aacs.js
tenacex: Changed.

lib/AACStream/aacs.js
tenacex: Changed.

tenacex:
Writing now.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+ * All rights reserved.
+ *
+ * A stream-based aac to mp4 converter. This utility can be used to
+ * deliver mp4s to a SourceBuffer on platforms that support native
+ * Media Source Extensions.
+ */
+'use strict';
+var Stream = require('../utils/stream.js');
+
+var
```
dmlap: Either everything below this needs to be indented differently or you should declare things here and then define them below.

debug/index.html
```diff
@@ -239,6 +239,7 @@ <h1 class="title">Transmux Analyzer</h1>
       codecsArray = codecs.value.split(',');
 
       mediaSource.addEventListener('sourceopen', function () {
+        mediaSource.duration = 0;
```
dmlap: Why set the duration to zero here?

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+
+ AacStream = function() {
+  var
+    everything,
+    receivedTimeStamp = false,
+    timeStamp = 0;
+
+  AacStream.prototype.init.call(this);
+
+  this.setTimestamp = function (frame) {
```
dmlap: Calling this method setTimestamp() implies some behaviors that wouldn't be true in this case. It's a common convention to create "setters" that exist to allow downstream developers to mutate some internal properties of an object. In that case though, the setter would normally take as an argument the value that is being set. I think it's a bit confusing to call this setTimestamp() and have it take a frame, given that practice.

How about splitting the side-effect portion of this out from the calculate-timestamp-from-frame portion? Something like:
```js
this.calculateTimestamp_ = function(frame) {
  /* ... */ 
};

// later, when you need to set the timestamp:
timestamp = this.calculateTimestamp_(frame);
```

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+    var d = frame.data,
+      size = ((d[3] & 0x01)  << 30) | (d[4]  << 22) | (d[5] << 14) | (d[6] << 6) | d[7] >>> 2;
+    size *= 4;
+    size += d[7] & 0x03;
+    timeStamp = size;
+  };
+
+  this.parseID3 = function(everything) {
```
dmlap: Shadowing `everything` here may confuse some people.

lib/AACStream/aacs.js
```diff
@@ -0,0 +1,120 @@
...
+    var returnSize = (everything[6] << 21) | (everything[7] << 14) |
+              (everything[8] << 7) | (everything[9]),
+      flags = everything[5],
+      footerpresent = (flags & 16) >> 4;
```
dmlap: variablesShouldBeCamelCased

debug/index.html
tenacex: This allows the vide element to calculate the duration itself.

lib/AACStream/aacs.js
tenacex: Fixed. I decided to declare the AacStream above and then keep the indentation the same below.

lib/AACStream/aacs.js
tenacex: The thing is that it is taking the value that is being set. It receives an ID3 tag that I then use to set the `timestamp` variable. Then whenever we emit an ADTS frame we check the timestamp value and add in the necessary info. It is setup like this because we can have an arbitrary amount of ADTS frames following an ID3 tag. I think we should discuss alternatives in person.

tenacex: Changed. See commit message.

lib/AACStream/aacs.js
tenacex: Changed to `header`

lib/AACStream/aacs.js
tenacex: iChangedThis.

debug/index.html
imbcmdth: After the append, the duration will be updated by MSE if it is greater than the existing duration.

imbcmdth:
There are still changes that need to be made here:
* Don't include `debug/mse-workbench.html` in this PR
* Rename `lib/AACStream` directory to just `lib/aac`
* Remove `lib/AACStream/metadata-stream.js`
* Remove `lib/AACStream/stream-types.js`
* Rename `lib/AACStream/aacs.js` to `lib/aac/index.js`
* Remove `test/files/trim.aac`

tenacex:
@imbcmdth & @dmlap I believe I have addressed most of your comment on this PR. Let me know if you have any more adjustments that you want made, and if not I can squash these commits and open up a new PR.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+    receivedTimeStamp = false,
+    timeStamp = 0;
+
+  AacStream.prototype.init.call(this);
+
+  this.setTimestamp = function (timestamp) {
+    timeStamp = timestamp;
+  };
+
+  this.parseID3 = function(header) {
```
imbcmdth: I think a name such as `parseID3TagSize` might be a more accurate name.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+              (header[8] << 7) | (header[9]),
+      flags = header[5],
+      footerPresent = (flags & 16) >> 4;
+    if (footerPresent) {
+            return returnSize + 20;
+    }
+    return returnSize + 10;
+  };
+
+  this.parseADTS = function(header) {
```
imbcmdth: I think a name such as `parseAdtsSize` might be a more accurate name.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+          chunk = {
+            type: 'timed-metadata',
+            data: everything.subarray(0, frameSize)
+          };
+
+          this.trigger('data', chunk);
+          everything = everything.subarray(frameSize);
+          continue;
+        } else {
+          break;
```
imbcmdth: You can flip the condition above and make this shorter:

````javascript
if (frameSize > everything.length) {
  break;
}
// ...everything else..
```

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+            type: 'audio',
+            data: everything.subarray(0, frameSize),
+            pts: timeStamp,
+            dts: timeStamp,
+          };
+          this.trigger('data', packet);
+          everything = everything.subarray(frameSize);
+          continue;
+        } else {
+          break;
```
imbcmdth: You can flip the condition above here too.

tenacex: Changed.

lib/aac/index.js
tenacex: Changed.

lib/aac/index.js
tenacex: Changed.

lib/aac/index.js
tenacex: Changed.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+      }
+      break;
```
imbcmdth: This is strange. It says, "we aren't in an ID3 tag or a ADTS frame so break and wait for more data." We try to be more resilient to errors than this. In my opinion, we should walk forward a byte and see if find a valid id3 tag or adts frame there.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+  };
+};
+
+AacStream.prototype = new Stream();
+
+var AACStream = {
+  AacStream: AacStream,
+  MetadataStream: require('../m2ts/metadata-stream'),
```
imbcmdth: There is no reason to require this and re-export it.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+        }
+      }
+      break;
+    }
+  };
+};
+
+AacStream.prototype = new Stream();
+
+var AACStream = {
```
imbcmdth: If you get rid of the MetadataStream export, you can get rid of this whole object and just export `AacStream` directly.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    while (everything.length >= 10) {
+      if ((everything[0] === 'I'.charCodeAt(0)) &&
+        (everything[1] === 'D'.charCodeAt(0)) &&
+        (everything[2] === '3'.charCodeAt(0))) {
+
+        //check framesize
+        frameSize = this.parseID3(everything);
+
```
imbcmdth: Extra line.

lib/aac/index.js
```diff
@@ -0,0 +1,121 @@
...
+
```
imbcmdth: Extra blank line.

lib/aac/index.js
tenacex: Changed. I removed the break and set `everything

lib/m2ts/metadata-stream.js
```diff
@@ -74,11 +74,14 @@ var
...
           // parse the description and URL fields
           tag.owner = parseIso88591(tag.data, 0, i);
+
           break;
         }
       }
+
       tag.privateData = tag.data.subarray(i + 1);
       tag.data = tag.privateData;
+
```
imbcmdth: These three changes are just added newlines. Remove them.

lib/m2ts/metadata-stream.js
```diff
@@ -113,8 +116,8 @@ MetadataStream = function(options) {
   }
 
   this.push = function(chunk) {
-    var tag, frameStart, frameSize, frame, i;
```
imbcmdth: This is also simply an extra newline.

lib/m2ts/metadata-stream.js
```diff
@@ -194,17 +198,27 @@ MetadataStream = function(options) {
       if (frameSize < 1) {
         return console.log('Malformed ID3 frame encountered. Skipping metadata parsing.');
       }
-
-      frame = {
-        id: String.fromCharCode(tag.data[frameStart],
+      var frameHeader = String.fromCharCode(tag.data[frameStart],
```
imbcmdth: Once again: only define variables at the top of files.

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
     audioTrack,
     packetStream, parseStream, elementaryStream,
-    aacStream, h264Stream,
+    adtsStream, h264Stream,aacStream,
     videoSegmentStream, audioSegmentStream, captionStream,
-    coalesceStream;
+    coalesceStream,
+    aacfile;
+
+  this.setupAACPipeline = function() {
```
imbcmdth: Our caps standard for this project means that this should be called `setupAacPipeline`.

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
+      aacStream.setTimestamp(frame.timestamp);
+    });
+    this.addAACListener();
+  };
 
-  Transmuxer.prototype.init.call(this);
-  options = options || {};
 
-  this.baseMediaDecodeTime = options.baseMediaDecodeTime || 0;
+  this.addAACListener = function() {
```
imbcmdth: Our caps standard for this project means that this should be called `addAacListener `.

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
-    .pipe(aacStream);
-
-  elementaryStream
-    .pipe(this.metadataStream)
-    .pipe(coalesceStream);
-
-  // Hook up CEA-608/708 caption stream
-  h264Stream.pipe(captionStream)
-    .pipe(coalesceStream);
+    this.setupTSPipeline = function() {
```
imbcmdth: Our caps standard for this project means that this should be called `setupTsPipeline`

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
-
-      // scan the tracks listed in the metadata
-      while (i--) {
-        if (!videoTrack && data.tracks[i].type === 'video') {
-          videoTrack = data.tracks[i];
-          videoTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
-        } else if (!audioTrack && data.tracks[i].type === 'audio') {
-          audioTrack = data.tracks[i];
-          audioTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
+  this.addTSListener = function() {
```
imbcmdth: Our caps standard for this project means that this should be called `addTsListener`

lib/mp4/transmuxer.js
imbcmdth: It would only be used once if you followed my two recommendations in the comments following this one. Also it should definitely have a name like `aacInput`

test/transmuxer.test.js
```diff
@@ -2484,3 +2488,59 @@ QUnit.test('buffers video samples until flushed', function() {
   QUnit.equal(segments[0].tags.audioTags.length, 0, 'generated no audio tags');
   QUnit.equal(segments[0].tags.videoTags.length, 2, 'generated two video tags');
 });
+
+QUnit.test('AAcStream parses correct ID3 tag size', function() {
```
imbcmdth: All of these tests should be in a `QUnit.module` named "AAC Stream". Then you can remove "AAcStream" from the test names.

lib/aac/index.js
tenacex: Changed.

lib/aac/index.js
tenacex: Changed.

lib/m2ts/metadata-stream.js
tenacex: Changed.

lib/m2ts/metadata-stream.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Changed.

lib/mp4/transmuxer.js
tenacex: Just took care of all of this.

debug/index.html
dmlap: Setting it to zero is not truthful and could mislead downstream clients. This doesn't seem correct to me.

dmlap: ... keep forgetting this is the debug page. Less important but it still seems like it could mislead *us* at some point.

lib/aac/stream-types.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = {
```
BrandonOCasey: Not sure if we need this file here, as it should be the same as m2ts stream-types

lib/m2ts/metadata-stream.js
```diff
@@ -194,17 +194,27 @@ MetadataStream = function(options) {
...
         return console.log('Malformed ID3 frame encountered. Skipping metadata parsing.');
       }
-
-      frame = {
-        id: String.fromCharCode(tag.data[frameStart],
+      frameHeader = String.fromCharCode(tag.data[frameStart],
                                 tag.data[frameStart + 1],
                                 tag.data[frameStart + 2],
-                                tag.data[frameStart + 3]),
+                                tag.data[frameStart + 3]);
```
BrandonOCasey: tag.data should all be aligned

lib/aac/stream-types.js
tenacex: Removed.

lib/m2ts/metadata-stream.js
tenacex: Changed.

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
-    .pipe(aacStream);
-
-  elementaryStream
-    .pipe(this.metadataStream)
-    .pipe(coalesceStream);
-
-  // Hook up CEA-608/708 caption stream
-  h264Stream.pipe(captionStream)
-    .pipe(coalesceStream);
+    this.setupTsPipeline = function() {
```
imbcmdth: This is too deeply indented.

lib/mp4/transmuxer.js
```diff
@@ -546,104 +547,170 @@ Transmuxer = function(options) {
...
+  } else {
+    this.setupAacPipeline();
+  }
+
+  // hook up the segment streams once track metadata is delivered
+
+
+
+
+
```
imbcmdth: Why so many empty lines?

PR #61, opened 2016-02-11
Added support for ignoring unsupported channels

tenacex:
The muxer now will not attempt to parse captions that occur on unsupported channels. Previously we were trying to parse multiple channels into one output stream, which caused issues. 

To do:
Add support for currently unsupported 608 caption commands.


imbcmdth:
Incorporated into #64

PR #63, opened 2016-02-16
Eighth note support

tenacex:
This patch supports the use of the eighth note special character in basic 608 captions.

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -213,7 +213,8 @@ var BASIC_CHARACTER_TRANSLATION = {
   0x7c: 0xf7,
   0x7d: 0xd1,
   0x7e: 0xf1,
-  0x7f: 0x2588
+  0x7f: 0x2588,
+
 };
 
 // Constants for the byte codes recognized by Cea608Stream. This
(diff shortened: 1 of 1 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -213,7 +213,7 @@ var BASIC_CHARACTER_TRANSLATION = {
   0x7c: 0xf7,
   0x7d: 0xd1,
   0x7e: 0xf1,
-  0x7f: 0x2588
+  0x7f: 0x2588,
```
imbcmdth: Extra comma snuck in there!

lib/m2ts/caption-stream.js
```diff
@@ -344,6 +344,14 @@ var Cea608Stream = function() {
         char0 = char1 = 0x20;
       }
 
+      // Look for special character sets
+      if ((char0 === 0x11 || char0 === 0x19) &&
+          (char1 >= 0x30 && char1 <= 0x3F)) {
+        // Put in eigth note and space
+        char0 = 0xE299AA;
+        char1 = 0x20;
```
imbcmdth: Does an eighth note always require a space afterwards?

tenacex: Changed.

lib/m2ts/caption-stream.js
tenacex: removed.

PR #64, opened 2016-02-22
Add channel test and fix flash

BrandonOCasey:
* added a getCharCode function to put the code in a central location
* only insert one space for codes that we don't support
* trim spaces before and after a row of captions before we emit
* ignore caption channels other than channel one
* updated tests to use a channel (previously non-existent)
* Added a test to decode multi channel captions (we only do the first channel)
* Verify that config and track are defined when we go to finish an FLV video frame

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -217,7 +217,7 @@ var BASIC_CHARACTER_TRANSLATION = {
 };
 
 var getCharFromCode = function(code) {
-  if(!code) {
+  if(code === null) {
     return '';
   }
   code = BASIC_CHARACTER_TRANSLATION[code] || code;
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -264,6 +264,10 @@ var Cea608Stream = function() {
   this.lastControlCode_ = null;
 
   this.push = function(packet) {
+    // Ignore other channels
+    if (packet['type'] !== 0) {
+      return;
+    }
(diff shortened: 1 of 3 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -216,6 +216,14 @@ var BASIC_CHARACTER_TRANSLATION = {
   0x7f: 0x2588
 };
 
+var getCharFromCode = function(code) {
+  if(!code) {
```
imbcmdth: This would also be falsey for code `0` which David thinks might be important one day. I would make this explicity `code !== null`

test/caption-stream.test.js
```diff
@@ -725,3 +740,28 @@ QUnit.skip('paint-on display mode', function() {
   QUnit.ok(false, 'not implemented');
 });
 
+QUnit.test('segment with multiple caption channels, we only parse 0', function() {
```
imbcmdth: Nice!

test/caption-stream.test.js
```diff
@@ -5,7 +5,8 @@ var
   m2ts = require('../lib/m2ts'),
   mp4 = require('../lib/mp4'),
   QUnit = require('qunit'),
-  sintelCaptions = require('./utils/sintel-captions');
+  sintelCaptions = require('./utils/sintel-captions'),
+  multichannelCaptions = require('./utils/multi-channel-captions');
```
imbcmdth: Can we trim this down? 10+mb is a little excessive for a test fixture!

PR #67, opened 2016-02-29
Restore segment buffering

dmlap:
As segments are produced by the transmuxer, we need to store them in-memory until the input is flushed. This was accidentally deleted in an earlier patch and caused the debug page to never show any output.


imbcmdth:
LGTM

PR #68, opened 2016-03-11
If Uint8Array is undefined (IE8) don't set up MP4 generator functions

gesinger:
Tested by ensuring that "'Uint8Array' is undefined" error is no longer being thrown in IE8, however, due to other errors, was not able to test full flow through with IE8.


BrandonOCasey:
We could probably use a comment and or error logging here, just to be safe.

gesinger:
Good call. Added a comment.

BrandonOCasey:
LGTM

PR #70, opened 2016-03-16
Fix exception when frames.length isz ero

surfdude75:


mrocajr:
LGTM

lib/mp4/transmuxer.js
```diff
@@ -503,9 +503,9 @@ VideoSegmentStream = function(track) {
 
     // For the last frame, use the duration of the previous frame if we
     // have nothing better to go on
-    if (frames.length &&
+    if (frames.length && (
```
forbesjo: Could you move this paren to the next line so that this can match the rest of the code in our plugins.

forbesjo:
I'll make a new PR with this fix and add that one little change.

imbcmdth:
Fixed in #90

PR #72, opened 2016-03-24
New aacfile docs

tenacex:
I added more documentation to the readme to explain how to set up a transmuxer that takes an aac file as input.


imbcmdth:
Removed the `aacfile` option in #76.

PR #76, opened 2016-03-31
Have the transmuxer pipeline automatically configure itself based on input

imbcmdth:
Experimental attempt to have the transmuxer pipeline get smarter about what to do with the data that is passed into it. 

* Added the ability to tear down the pipeline
* Added a sniffer function that checks to see if the data is likely an AAC by checking to see if it starts with an ID3 tag
* The pipeline creation now happens when the first segment or when a different type of segment is passed in
* The audio track is saved so that the `timelineStartInfo` data is preserved between pipeline changes

### DIFF
--- a/lib/aac/index.js
+++ b/lib/aac/index.js
@@ -20,7 +20,7 @@ var AacStream;
 
 AacStream = function() {
   var
-    everything,
+    everything = new Uint8Array(),
     receivedTimeStamp = false,
     timeStamp = 0;
 
@@ -65,7 +65,7 @@ AacStream = function() {
 
     // If there are bytes remaining from the last segment, prepend them to the
     // bytes that were pushed in
-    if (everything && everything.length) {
+    if (everything.length) {
       tempLength = everything.length;
       everything = new Uint8Array(bytes.byteLength + tempLength);
       everything.set(everything.subarray(0, tempLength));
@@ -134,7 +134,7 @@ AacStream = function() {
     if (bytesLeft > 0) {
       everything = everything.subarray(byteIndex);
     } else {
-      everything = null;
+      everything = new Uint8Array();
     }
   };
 };
--- a/lib/aac/index.js
+++ b/lib/aac/index.js
@@ -58,13 +58,14 @@ AacStream = function() {
     var
       frameSize = 0,
       byteIndex = 0,
+      bytesLeft,
       chunk,
       packet,
       tempLength;
 
     // If there are bytes remaining from the last segment, prepend them to the
     // bytes that were pushed in
-    if (everything !== undefined && everything.length) {
+    if (everything && everything.length) {
       tempLength = everything.length;
       everything = new Uint8Array(bytes.byteLength + tempLength);
       everything.set(everything.subarray(0, tempLength));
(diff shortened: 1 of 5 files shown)


lib/aac/index.js
```diff
@@ -110,6 +129,13 @@ AacStream = function() {
       }
       byteIndex++;
     }
+    bytesLeft = everything.length - byteIndex;
+
+    if (bytesLeft > 0) {
+      everything = everything.subarray(byteIndex);
+    } else {
+      everything = null;
```
dmlap: Not important but I'd lean towards making `everything` an empty typed array when there's no data to process and removing the need to check for nulls.

lib/mp4/transmuxer.js
```diff
@@ -929,20 +939,44 @@ CoalesceStream.prototype.flush = function(flushSource) {
...
+    emitData,
+    emitDone;
 
   Transmuxer.prototype.init.call(this);
 
   options = options || {};
   this.baseMediaDecodeTime = options.baseMediaDecodeTime || 0;
+  this.currentPipeline_ = null;
+
+  this.teardownAacPipeline = function() {
```
dmlap: Instead of these teardown functions, what about just tossing out the old pipeline and creating a new one?

lib/utils/stream.js
```diff
@@ -12,16 +12,19 @@
 var Stream = function() {
   this.init = function() {
     var listeners = {};
+    var sentinels = {};
     /**
      * Add a listener for a specified event type.
      * @param type {string} the event name
      * @param listener {function} the callback to be invoked when an event of
      * the specified type occurs
```
BrandonOCasey: maybe add info on the sentinel variable in the doc comments (you can keep the same reversed style if you want, since it will keep it consistent even though its technically wrong)

lib/utils/stream.js
```diff
@@ -92,11 +102,18 @@ var Stream = function() {
...
 
   this.on('done', function(flushSource) {
     destination.flush(flushSource);
-  });
+  }, destination);
+
+  return destination;
+};
+
+Stream.prototype.unpipe = function(destination) {
```
BrandonOCasey: if we can just throw away the pipeline as David suggested, than we can probably get rid of unpipe

lib/mp4/transmuxer.js
```diff
@@ -1100,23 +1180,39 @@ Transmuxer = function(options) {
...
   // flush any buffered data
   this.flush = function() {
+    hasFlushed = true;
     // Start at the top of the pipeline and flush all pending work
     headOfPipeline.flush();
   };
 
-  // Re-emit any data coming from the coalesce stream to the outside world
-  coalesceStream.on('data', function (data) {
+  emitData = function (data) {
```
dmlap: I'm not really a fan of these helper functions for single method calls. Totally subjective, I know.

lib/utils/stream.js
```diff
@@ -12,16 +12,19 @@
...
     var listeners = {};
+    var sentinels = {};
     /**
      * Add a listener for a specified event type.
      * @param type {string} the event name
      * @param listener {function} the callback to be invoked when an event of
      * the specified type occurs
      */
-    this.on = function(type, listener) {
+    this.on = function(type, listener, sentinel) {
```
dmlap: I think we should try very hard to stick with a node-compatible API here. First, we keep discussing using browserify's Stream implementation and this would make that difficult. Secondly, it may be confusing to people learning the project to see something that looks *really really* like a Stream but has a subtly different API.

lib/mp4/transmuxer.js
```diff
@@ -1100,23 +1180,39 @@ Transmuxer = function(options) {
 
   // feed incoming data to the front of the parsing pipeline
   this.push = function(data) {
+    if (hasFlushed) {
```
dmlap: If we don't support dynamic reconfiguration of the transmuxer after flushing, most of this code wouldn't be necessary. What is the use-case for reconfiguring the transmuxer after flushes?

imbcmdth: It allows us to support, seamlessly, situations where we have audio in both (TS; AAC) formats in one manifest. Might as well support that now and not worry about it in the future.

lib/utils/stream.js
imbcmdth: The sentinel is used by the pipe function internally. Externally, the `pipe`-related api is exactly the same. We can make it so that the "event" api is the same by creating internal-only event binding functions and exposing them via a two-argument proxy. I feel that may be more work for little benefit.

lib/utils/stream.js
imbcmdth: Stupid reversed doc comments!

dmlap:
LGTM

PR #80, opened 2016-04-05
add audio track info to event that are sent to MSE

BrandonOCasey:
Tested and ready to review!

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -34,7 +34,6 @@ var videoProperties = [
   'profileCompatibility',
 ];
 
-
 // object types
 var VideoSegmentStream, AudioSegmentStream, Transmuxer, CoalesceStream;
 
@@ -131,7 +130,7 @@ AudioSegmentStream = function(track) {
 
     if (track) {
       audioProperties.forEach(function(prop) {
-        track[prop] = data[prop] || '';
+        track[prop] = data[prop];
       });
     }
 
@@ -272,7 +271,7 @@ VideoSegmentStream = function(track) {
       track.sps = [nalUnit.data];
 
       videoProperties.forEach(function(prop) {
-        track[prop] = config[prop] || '';
+        track[prop] = config[prop];
       }, this);
     }
 
@@ -875,12 +874,12 @@ CoalesceStream.prototype.flush = function(flushSource) {
   if (this.videoTrack) {
     timelineStartPts = this.videoTrack.timelineStartInfo.pts;
     videoProperties.forEach(function(prop) {
-      event.info[prop] = this.videoTrack[prop] || '';
+      event.info[prop] = this.videoTrack[prop];
     }, this);
   } else if (this.audioTrack) {
     timelineStartPts = this.audioTrack.timelineStartInfo.pts;
     audioProperties.forEach(function(prop) {
-      event.info[prop] = this.audioTrack[prop] || '';
+      event.info[prop] = this.audioTrack[prop];
     }, this);
   }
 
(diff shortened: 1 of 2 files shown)


lib/mp4/transmuxer.js
```diff
@@ -862,6 +864,14 @@ CoalesceStream.prototype.flush = function(flushSource) {
     timelineStartPts = this.videoTrack.timelineStartInfo.pts;
   } else if (this.audioTrack) {
     timelineStartPts = this.audioTrack.timelineStartInfo.pts;
+    ['audioobjecttype',
```
imbcmdth: I would love it if we could camelCase these names but that would be a more major change so I'm just complaining and don't expect you to change that for this PR

lib/mp4/transmuxer.js
```diff
@@ -862,6 +864,14 @@ CoalesceStream.prototype.flush = function(flushSource) {
...
+     'channelcount',
+     'samplerate',
+     'samplingfrequencyindex',
+     'samplesize'].forEach(function(audioProp) {
+       event.info[audioProp] = self.audioTrack[audioProp];
+     });
```
imbcmdth: `forEach` takes a second parameter as the context the function is given when executed.
```
[...].forEach(function(audioProp) {
       event.info[audioProp] = this.audioTrack[audioProp];
     }, this);
```

lib/mp4/transmuxer.js
```diff
@@ -112,11 +130,9 @@ AudioSegmentStream = function(track) {
     collectDtsInfo(track, data);
 
     if (track) {
-      track.audioobjecttype = data.audioobjecttype;
-      track.channelcount = data.channelcount;
-      track.samplerate = data.samplerate;
-      track.samplingfrequencyindex = data.samplingfrequencyindex;
-      track.samplesize = data.samplesize;
+      audioProperties.forEach(function(prop) {
+        track[prop] = data[prop] || '';
```
gesinger: Will the empty string default change any of the prior behavior?

gesinger:
One question (as a comment), and do we want to have any tests to check that the properties are in there (besides just checking the info object)? Otherwise, LGTM.

lib/mp4/transmuxer.js
BrandonOCasey: previously we did not do anything with this information, and from what I have seen it is always defined. If it was not defined the muxer probably would not even be able to get this far. So we could probably forgo the default empty string if we wanted to.

BrandonOCasey:
I will have the tests check that the properties exist

gesinger:
LGTM

lib/mp4/transmuxer.js
```diff
@@ -1145,4 +1164,6 @@ module.exports = {
   Transmuxer: Transmuxer,
   VideoSegmentStream: VideoSegmentStream,
   AudioSegmentStream: AudioSegmentStream,
+  audioProperties: audioProperties,
+  videoProperties: videoProperties
```
dmlap: I'm not 100% consistent on it myself but I try to do ALL_CAPS_CASE for constants.

lib/mp4/transmuxer.js
```diff
@@ -822,7 +834,8 @@ CoalesceStream.prototype.flush = function(flushSource) {
     offset = 0,
     event = {
       captions: [],
-      metadata: []
+      metadata: [],
+      info: {}
```
dmlap: I kind of wish we had a more descriptive name for this field but it's not worth holding up the PR on.

dmlap: Another thought: we need to document these properties at some point.

BrandonOCasey: maybe "structure" or "properties"?

PR #84, opened 2016-04-08
Display roll-up caption rows as one cue to preserve ordering

forbesjo:
This PR is to fix an issue for roll-up captions where the rows would stack on top of each other.

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -382,18 +382,20 @@ Cea608Stream.prototype = new Stream();
 // Trigger a cue point that captures the current state of the
 // display buffer
 Cea608Stream.prototype.flushDisplayed = function(pts) {
-  var row, i;
+  var content = this.displayed_
+    // remove spaces from the start and end of the string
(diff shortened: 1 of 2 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -382,18 +382,20 @@ Cea608Stream.prototype = new Stream();
...
-      this.trigger('data', {
-        startPts: this.startPts_,
-        endPts: pts,
-        // remove spaces from the start and end of the string
-        text: row.trim()
-      });
-    }
+  var content = this.displayed_
+    // remove spaces from the start and end of the string
+    .map(function(row){ return row.trim(); })
```
imbcmdth: We don't have vjsstandard running on this project but if we did, it would complain about the missing space after the closing parens in the function definition.

imbcmdth:
Minor nit, otherwise LGTM.

lib/m2ts/caption-stream.js
forbesjo: I just noticed that too, fixing.

PR #86, opened 2016-04-14
Paint-on captions

tenacex:
Paint-on captions implemented within our existing system. Pending tests and an actual video with 608 captions, I tested using captions that were meant for rollup and forced paint-on mode and it worked well.


squarebracket:
Added paint-on support in #171. Closing.

PR #87, opened 2016-04-21
Fixed debug page when using muxed output

tenacex:
Since we changed some things around, the debug page wasn't appending to the buffer when we had the remux option checked. This fixes that. There also were some typos which were causing errors.


imbcmdth:
I think my latest commit fixed this in a slightly different way. I was using (re)muxed output without issue.

PR #88, opened 2016-04-25
Don't emit empty nals

tenacex:
Sometimes, due to a corrupted stream, we are emitting empty nal units, causing regression errors. This screens out empty nal events.

Example:

If we pass in a Nal unit such as:
```javascript
new Uint8Array([
    0x00, 0x00, 0x01,  // start code
    0x00, 0x00,           // empty payload
    0x00, 0x00, 0x00  // end code
]);
```
we would end up emitting it and adding an empty array to the array of nal units we feed through the pipeline. Stages of the pipeline rely on these arrays to not be empty, so it can sometimes break playback if we don't screen these out early in the pipeline.

### DIFF
--- a/lib/codecs/h264.js
+++ b/lib/codecs/h264.js
@@ -61,7 +61,7 @@ NalByteStream = function() {
         }
 
         // deliver the NAL unit
-        if (buffer.subarray(syncPoint + 3, i - 2).length > 0) {
+        if (syncPoint + 3 !==  i - 2) {
           this.trigger('data', buffer.subarray(syncPoint + 3, i - 2));
         }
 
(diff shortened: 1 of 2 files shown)


lib/codecs/h264.js
```diff
@@ -61,7 +61,9 @@ NalByteStream = function() {
         }
 
         // deliver the NAL unit
-        this.trigger('data', buffer.subarray(syncPoint + 3, i - 2));
+        if (buffer.subarray(syncPoint + 3, i - 2).length > 0) {
```
imbcmdth: Don't create a subarray just to check the length when you can just make sure that `syncPoint + 3 !== i - 2`!

dmlap:
Example? Test?

PR #94, opened 2016-05-25
Produce dists for FLV and MP4 outputs

dmlap:
Create three dist files: one with everything, one for just FLV output, and one for just MP4 output. Our uses of this project in video.js don't require support for all formats so we can save some code by having specialized builds. The primary dist (i.e. "mux.js") still includes everything.


lib/flv/index.js
```diff
@@ -1,5 +1,4 @@
 module.exports = {
   tag: require('./flv-tag'),
-  Transmuxer: require('./transmuxer'),
-  tools: require('../tools/flv-inspector'),
```
gesinger: Should this be added into lib/index under muxjs.flv.tools?

dmlap: Good point.

gesinger:
LGTM

PR #95, opened 2016-05-26
Linting

dmlap:
Since jshint was not even catching unused vars, replace it with eslint. Turn off linting for now since there are so many errors. Also, produce compressed dist files.


imbcmdth:
LGTM

PR #97, opened 2016-06-06
Remove creation of init segments and serialize track information

gesinger:
DO NOT MERGE

This will require a major version update. We might want it to exist on a branch after review.

Associated videojs-contrib-media-sources PR https://github.com/videojs/videojs-contrib-media-sources/pull/92


lib/mp4/transmuxer.js
```diff
@@ -891,17 +892,10 @@ CoalesceStream.prototype.flush = function(flushSource) {
 
   this.emittedTracks += this.pendingTracks.length;
 
-  initSegment = mp4.initSegment(this.pendingTracks);
+  event.serializedTracks = mp4Utils.serializeTracks(this.pendingTracks);
```
gesinger: This may not be the best method for passing the data out of the worker, but the data seemed small enough that it wouldn't make a big difference in performance, and was relatively straightforward. I'd definitely be open to alternative approaches though.

test/transmuxer.test.js
```diff
@@ -2515,64 +2511,6 @@ QUnit.test('pipeline dynamically configures itself based on input', function() {
   QUnit.equal(transmuxer.transmuxPipeline_.type, 'aac', 'detected AAC file data');
 });
 
-QUnit.test('reuses audio track object when the pipeline reconfigures itself', function() {
```
gesinger: Now that the timing information (specifically, baseMediaDecodeTime) is going to be handled at a higher level, I am working on adding the test into media-sources.

lib/mp4/transmuxer.js
imbcmdth: There are two options here.
1. Instead of manually serializing the whole track, you can simply convert the `sps` and `pps` into arrays of arrays and then convert them back to arrays of typed-arrays on the other size of the worker.
1. The entries in the `sps` and `pps` can be duplicated in the worker and sent via a zero-copy out of the worker much like the `data` is..

Either way, I don't think the muxer is the place to do the conversion since it shouldn't know or care if it is in a typed array or not. The place for this kind of `data-marshalling` is in the `transmux_worker.js` file in `videojs-contrib-media-sources`.

imbcmdth:
Also at some point, you will want to move the baseMediaDecodeTime stuff also.

lib/mp4/transmuxer.js
gesinger: You're definitely right that any manipulation that is done for dealing with the web worker should be in 'transmux-worker' in videojs-contrib-media-sources. I moved it there.

Right now I'm still doing the objToTypedArr and typedArrToObj in videojs-contrib-media-sources, as I still wanted the other information from the tracks object (except samples, which I delete to save space on the copy), and it seemed simpler than doing the zero copy for multiple conditionally existing inner sps and pps values (re: 2), and retains the offset values (re: 1). Though I could be wrong on which is simpler/more straightforward. Let me know what you think on that PR!

imbcmdth:
Superseded by other changes

PR #98, opened 2016-06-10
Added "isNonSyncSample" missing flag and "startTime" event detail

vpalmisano:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mux.js",
-  "version": "2.3.0",
+  "version": "2.3.1",
```
gkatsev: versioning is handled when we're ready for release, so, a PR shouldn't be updating it.

dmlap:
@vpalmisano thanks for the PR! Can you provide us with the use-case you're trying to address with these new properties?

imbcmdth:
@dmlap It seems that `isNonSyncSample` is used in MP4 to specify samples that can not be decoded independently. Not sure what benefit marking them has but it doesn't seem wrong to set that flag properly. I would love to find out from @vpalmisano what the purpose really is inside of MSE!

imbcmdth:
Intersting comment in Chromium source implies that the flag is not important there since we do set depends_on flag correctly https://cs.chromium.org/chromium/src/media/formats/mp4/track_run_iterator.cc?sq=package:chromium&l=167

forbesjo:
We have this feature now, thank you for the contribution. Closing

PR #103, opened 2016-07-20
MP4 Probe

dmlap:
- Add support for v2 tfdt
Correctly parse tfdts with 64-bit fields. We can't avoid the possibility of overflow but values that aren't too big for js numbers should work.

- Utility for probing start times from MP4s
Add functions to extract track timescales and base media decode times from fragmented MP4s. This is intended to be used to correctly set the timestampOffset when playing back MP4s that have a non-zero base media decode time to start with. Re-arrange some test utility functions to avoid duplicating sample test data.

### DIFF
--- a/test/mp4-inspector.test.js
+++ b/test/mp4-inspector.test.js
@@ -21,53 +21,11 @@
 */
 var
   mp4 = require('../lib/mp4'),
+  mp4Helpers = require('./utils/mp4-helpers'),
   QUnit = require('qunit'),
-  typeBytes = function(type) {
-    return [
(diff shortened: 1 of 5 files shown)


test/mp4-inspector.test.js
```diff
@@ -21,53 +21,11 @@
 */
 var
   mp4 = require('../lib/mp4'),
+  mp4Helpers = require('./utils/mp4-helpers'),
```
imbcmdth: Nice!

imbcmdth:
LGTM

PR #104, opened 2016-07-28
Timestamp Rollover

mjneil:
## Description
Every 26 hours the PTS value resets to 0. We need to detect this condition in the Muxer and add a value to make the PTS values continuous.
* Rollover conditions:
  * Between a single PES packet's DTS and it's PTS
  * Between a PES packet's PTS and the next packet's DTS
* There is a spec for M2TS playback over MSE  https://w3c.github.io/media-source/mp2t-byte-stream-format.html#mpeg2ts-discontinuities

## Specific Changes proposed
* Detect when the PTS OR DTS value rolls over
* Adjust PTS and DTS values to be always increasing by adding `n*2^33` where `n` is the number of rollovers.
* Attach a `TSCorrectionStream` to the end of the `ElementaryStream` or `AacStream` in the pipeline to make these adjustments
* Tests

## Requirements Checklist
- [ ] Feature implemented / Bug fixed
- [ ] Unit Tests updated or fixed

### DIFF
--- a/lib/flv/transmuxer.js
+++ b/lib/flv/transmuxer.js
@@ -403,7 +403,8 @@ Transmuxer = function(options) {
     self = this,
 
     packetStream, parseStream, elementaryStream,
-    videoCorrectionStream, audioCorrectionStream, timedMetadataCorrectionStream,
+    videoTimestampRolloverStream, audioTimestampRolloverStream,
+    timedMetadataTimestampRolloverStream,
     adtsStream, h264Stream,
     videoSegmentStream, audioSegmentStream, captionStream,
     coalesceStream;
@@ -421,9 +422,9 @@ Transmuxer = function(options) {
   packetStream = new m2ts.TransportPacketStream();
   parseStream = new m2ts.TransportParseStream();
   elementaryStream = new m2ts.ElementaryStream();
-  videoCorrectionStream = new m2ts.TSCorrectionStream('video');
-  audioCorrectionStream = new m2ts.TSCorrectionStream('audio');
-  timedMetadataCorrectionStream = new m2ts.TSCorrectionStream('timed-metadata');
+  videoTimestampRolloverStream = new m2ts.TimestampRolloverStream('video');
+  audioTimestampRolloverStream = new m2ts.TimestampRolloverStream('audio');
+  timedMetadataTimestampRolloverStream = new m2ts.TimestampRolloverStream('timed-metadata');
 
   adtsStream = new AdtsStream();
   h264Stream = new H264Stream();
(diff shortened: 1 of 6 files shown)


lib/m2ts/ts-correction-stream.js
```diff
@@ -0,0 +1,88 @@
...
+var Stream = require('../utils/stream');
+
+var MAX_TS = 8589934592;
+
+var RO_THRESH = 4294967296;
+
+var TSCorrectionStream = function() {
+
+  var
+    reference = {
```
imbcmdth: Do you think this code could be somewhat simplified if the `TSCorrectionStream` took a parameter on in the constructor to specify the stream type that is handled and ignored packets from every other `type` in the push call?

lib/m2ts/ts-correction-stream.js
```diff
@@ -0,0 +1,88 @@
...
+        referenceDTS: undefined
+      }
+    };
+
+  TSCorrectionStream.prototype.init.call(this);
+
+  var handleRollover = function(value, reference) {
+    var direction = 1;
+
+    if (value > reference) {
```
imbcmdth: I think this might need a comment explaining the idea behind this logic and why `value > reference` means `direction = -1`. It took me a few moments to get it and I more or less knew what to expect!

lib/m2ts/ts-correction-stream.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    this.trigger('done');
+  };
+
+};
+
+TSCorrectionStream.prototype = new Stream();
+
+module.exports = TSCorrectionStream;
```
imbcmdth: Missing a new line! :wink:

test/transmuxer.test.js
```diff
@@ -877,6 +886,104 @@ QUnit.test('drops packets with unknown stream types', function() {
...
+  QUnit.equal(packets[2].pts, Math.pow(2, 33), 'correctly parses the max pts value');
+  QUnit.equal(packets[3].pts, Math.pow(2, 33) + 50, 'correctly parsed the rollover pts value');
+});
+
+QUnit.test('Correctly parses multiple PTS rollovers', function() {
+  var
+    packets = [],
+    packetData = [0x01, 0x02],
+    pesArray = [pesHeader(false, 1),
+                pesHeader(false, Math.floor(Math.pow(2, 33) * (1 / 3))),
```
imbcmdth: Might be nice to simply make a constant for the value 2^33 instead of repeating it or even a function so you can simply do `pesHeader(false, makeTS(1 / 3))`?

lib/m2ts/ts-correction-stream.js
mjneil: It depends on where we want the complexity. The current proposal allows us to only have a single `TSCorrectionStream` in the `Transmuxer`. The other option would require the transmuxer to create multiple streams (e.g. 3 different streams in the ts pipeline).

lib/flv/transmuxer.js
```diff
@@ -420,6 +421,10 @@ Transmuxer = function(options) {
   packetStream = new m2ts.TransportPacketStream();
   parseStream = new m2ts.TransportParseStream();
   elementaryStream = new m2ts.ElementaryStream();
+  videoCorrectionStream = new m2ts.TSCorrectionStream('video');
```
imbcmdth: Thinking about this some more, a good name for the stream stage is `TimestampRolloverStream` and then `videoTimestampRolloverStream` et.al. which better describes what this stage does.

Issue #100, opened 2016-06-29
Sorting of pts causes Chrome to display incorrect CEA-608 embedded captions


ianwelsh:
I was having this issue, https://github.com/videojs/videojs-contrib-hls/issues/761.

But when I remove the following code from line 190 of caption_stream.js:

```
this.captionPackets_.sort(function(a, b) {
    return a.pts - b.pts;
  });
```

The problem resolves itself. Tested in Chrome, Firefox, and Safari. 

Can someone explain why this code is necessary? Is this code OK to take out without breaking something else?
If not, are there any workarounds I can implement?

ianwelsh:
I believe this is due to Chrome having an unstable Array.prototype.sort() implementation (https://bugs.chromium.org/p/v8/issues/detail?id=90), whereas Firefox and Safari have stable implementations.

### Example
**Pre-sort:**
```
[{"type":0,"pts":8353875960,"ccData":8320},
{"type":0,"pts":8353875960,"ccData":37925},
{"type":0,"pts":8353875960,"ccData":38061}]
```
**Post-sort in Chrome:**
```
[{"type":0,"pts":8353875960,"ccData":37925},
{"type":0,"pts":8353875960,"ccData":38061},
{"type":0,"pts":8353875960,"ccData":8320}]
```
**Post-sort in Firefox:**
```
[{"type":0,"pts":8353875960,"ccData":8320},
{"type":0,"pts":8353875960,"ccData":37925},
{"type":0,"pts":8353875960,"ccData":38061}]
```

Any reason why I shouldn't take out the sort? Are there any conditions where the caption packets would get unsorted?

dmlap:
@ianwelsh if the video is encoded with b-frames, caption packets will arrive out-of-order and so it's important to resort them before display. Good find about the unstable sort() though! I suspect you're right and that is causing the problem. We should find some way to disambiguate our comparator so that there are no equivalent caption packets during the sorting.

PR #107, opened 2016-08-10
Fix: Force stable sort of cea608 captions

ianwelsh:
Add a presortIndex property to cea608 caption objects in order to disambiguate the sort so that we get a stable sort
See issue: https://github.com/videojs/mux.js/issues/100


imbcmdth:
Merged via #112 thanks for the fix @ianwelsh!!

PR #112, opened 2016-08-24
608 captions stable-sort for Chrome

### LINKED ISSUE #100: Sorting of pts causes Chrome to display incorrect CEA-608 embedded captions

imbcmdth:
Minor cleanup to #107 that fixes #100

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -186,14 +186,18 @@ CaptionStream.prototype.flush = function() {
     return;
   }
 
-  // map the presort index so we can use it to disambiguate the sort()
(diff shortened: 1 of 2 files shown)


forbesjo:
LGTM

PR #114, opened 2016-09-07
Add Bundle Collapser to save 159 gzipped bytes!

BrandonOCasey:
Also added an uglify build step for `dist/mux.js` to `dist/mux.min.js`


dmlap:
LGTM

forbesjo:
This looks good to me as well

PR #115, opened 2016-09-07
Set the PTS and DTS for TS timestamp ID3 tags

dmlap:
The special PRIV frame that specifies the timestamp in raw AAC data would not itself receive a timestamp since it was already parsed when the base timestamp was set. Add special case handling for that ID3 tag so that it does not end up with a cueTime of NaN when picked up by the CoalesceStream in the MP4 transmuxer.


mjneil:
LGTM

PR #118, opened 2016-09-19
Added support for generating version 1 TFDT boxes

imbcmdth:
Switch to using version 1 tfdt boxes to allow up to 64-bits for the baseMediaDecodeTime. 

Since we use a 90kHz clock for the BMDT, we can exceed a 32-bit value after ~13.256 hours of video. This can cause a problem with live-streams or extremely long VODs as the BMDT will wrap around and playback will likely break.

### DIFF
--- a/lib/mp4/mp4-generator.js
+++ b/lib/mp4/mp4-generator.js
@@ -536,8 +536,9 @@ tkhd = function(track) {
  * about tracks in a movie fragment (moof) box.
  */
 traf = function(track) {
-  var trackFragmentHeader, trackFragmentDecodeTime,
-      trackFragmentRun, sampleDependencyTable, dataOffset;
+  var trackFragmentHeader, trackFragmentDecodeTime, trackFragmentRun,
+      sampleDependencyTable, dataOffset, maxWordValue,
+      upperWordBaseMediaDecodeTime, lowerWordBaseMediaDecodeTime;
 
   trackFragmentHeader = box(types.tfhd, new Uint8Array([
     0x00, // version 0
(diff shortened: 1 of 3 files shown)


lib/mp4/mp4-generator.js
```diff
@@ -552,21 +553,29 @@ traf = function(track) {
     0x00, 0x00, 0x00, 0x00  // default_sample_flags
   ]));
 
+  maxWordValue = Math.pow(2, 32);
```
dmlap: This seems like it should be a constant.

test/transmuxer.test.js
```diff
@@ -1995,11 +1995,14 @@ QUnit.test('calculates baseMediaDecodeTime values from the first DTS ever seen a
 });
 
 QUnit.test('calculates baseMediaDecodeTime values relative to a customizable baseMediaDecodeTime', function() {
-  var segment, boxes, tfdt;
+  var segment, boxes, tfdt, bMDTValue;
+
+  bMDTValue = Math.pow(2, 32) + 100;
```
dmlap: I try to avoid acronyms unless they're better known than the full term.

test/transmuxer.test.js
```diff
@@ -2033,7 +2036,7 @@ QUnit.test('calculates baseMediaDecodeTime values relative to a customizable bas
 
   boxes = mp4.tools.inspect(segment);
   tfdt = boxes[0].boxes[1].boxes[1];
-  QUnit.equal(tfdt.baseMediaDecodeTime, 1324, 'calculated baseMediaDecodeTime');
+  QUnit.equal(tfdt.baseMediaDecodeTime, bMDTValue + 90, 'calculated baseMediaDecodeTime');
```
dmlap: The `+ 90` part is a little unclear to me. Maybe a comment about how that's calculated?

PR #120, opened 2016-09-21
Create mp4 init segment only for first segment of timeline

mjneil:
## Description
Currently we generate and attach an initialization segment (`ftyp` and `moov` boxes) to the mp4 we generate for every ts segment. Adding this init segment for each segment causes a noticeable hiccup in the audio along segment boundaries. Only attaching an init segment on timeline changes when `baseMediaDecodeTime` is set resolves this issue.

Requires [pull request](https://github.com/videojs/videojs-contrib-media-sources/pull/104)

## Specific Changes proposed
Only create a single mp4 init segment whenever `setBaseMediaDecodeTime` is called, otherwise just create the `moof` and `mdat`.

## Requirements Checklist
- [x] Feature implemented / Bug fixed
- [x] Unit Tests updated or fixed

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -791,10 +791,6 @@ CoalesceStream = function(options, metadataStream) {
   this.pendingMetadata = [];
   this.pendingBytes = 0;
   this.emittedTracks = 0;
-  this.createInitSegment = {
-    VideoSegmentStream: true,
-    AudioSegmentStream: true
-  };
 
   CoalesceStream.prototype.init.call(this);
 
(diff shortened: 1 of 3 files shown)


dmlap:
We should probably have a test that shows that the init segment is re-generated when base media decode time gets reset. Otherwise, LGTM

lib/mp4/transmuxer.js
```diff
@@ -887,15 +887,15 @@ CoalesceStream.prototype.flush = function(flushSource) {
...
   // Create an init segment containing a moov
   // and track definitions
-  event.data.set(initSegment);
-  offset += initSegment.byteLength;
+  event.initSegment.set(initSegment);
+
+  // Create a new typed array to hold the moof+mdat
+  event.data = new Uint8Array(this.pendingBytes);
 
   // Append each moof+mdat (one per track) after the init segment
```
gesinger: Minor, but this comment is a bit outdated from the change

test/transmuxer.test.js
```diff
@@ -2839,79 +2851,51 @@ QUnit.test('parses an example mp2t file and generates combined media segments',
...
             'generated identical audio trafs');
-  QUnit.equal(segments[0][4].boxes[0].sequenceNumber,
+  QUnit.equal(boxes[0][2].boxes[0].sequenceNumber,
         0,
         'set the correct audio sequence number');
-  QUnit.equal(segments[1][2].boxes[0].sequenceNumber,
+  QUnit.equal(boxes[1][2].boxes[0].sequenceNumber,
         1,
         'set the correct audio sequence number');
-  QUnit.deepEqual(segments[0][5],
```
gesinger: Is everything here and down no longer relevant?

mjneil: yes, its been removed.

lib/mp4/transmuxer.js
```diff
@@ -887,17 +887,17 @@ CoalesceStream.prototype.flush = function(flushSource) {
 
   initSegment = mp4.initSegment(this.pendingTracks);
 
-  // Create a new typed array large enough to hold the init
-  // segment and all tracks
-  event.data = new Uint8Array(initSegment.byteLength +
-                              this.pendingBytes);
+  // Create a new typed array to hold the init segment
+  event.initSegment = new Uint8Array(initSegment.byteLength);
```
gesinger: Is it worth it to try to pass down whether we want the initSegment created or not, or is having to maintain that state too costly/is the creation and passing of the init segment cheap enough that it isn't a big deal (or is there another concern I'm missing)?

mjneil: Originally I attempted to "pass down" whether we wanted to create the init segment by setting the flag whenever a discontinuity is encountered and `setBaseMediaDecodeTime` is called. However, we also need to add the init segment on audio configuration changes, which the transmuxer has no knowledge of. Since the transmuxer has been sending the init segment data along with every segment to the source buffers anyway, we are not really adding more overhead by continuing to send it over. I feel it simplifies things to let the transmuxer just handle transmuxing and leave the decision to media-sources

gesinger:
LGTM

imbcmdth:
LGTM

Issue #119, opened 2016-09-21
Event listeners adding/removing event listeners breaks trigger implementation


odinho:
Hello!

If you use an event from mux.js to (say) remove yourself from further events, because you had a one-time job, the code you have in place will misbehave in the for-loop going through all the callbacks.

I found a patch for this in our project: https://github.com/hola/mux.js/commit/f02fea0647e3540714df8e72940888d28b923b4a

I'd like to get it off our hands and into upstream (you).  An alternative route fixing this, and I can do this if you'd accept it, is to never mutate the current Array when you do on/off, but always make a copy of the Array to listeners.

That way using it will be safe, and there's no overhead in `trigger`, only in `on` and `off` which hopefully is called less often.

dmlap:
Either one would be *great* and we'd definitely appreciate the patch. I really like your proposal to do the copy-on-write in `on`/`off`-- `trigger` is a hot section so it would be great to avoid the allocation there.

PR #121, opened 2016-09-23
Allow add/remove listeners in callbacks (#119)

odinho:
Using a callback to mutate the listeners (e.g. removing yourself), previously
changed the iterated array, causing bugs.  Remove the problem by always copying
the array, so on()/off() become safe to call at any time.


odinho:
Also added a test, because why not.

dmlap:
Awesome. LGTM

imbcmdth:
LGTM

dmlap:
Thanks for the contribution, @odinho!

PR #122, opened 2016-09-23
add ptsTime to ts inspector output

mjneil:
This changes the output of the ts inspector to include both `pts` and `ptsTime`. `pts` is the timestamp value on the 90khz clock and `ptsTime` is the timestamp value converted into seconds.


imbcmdth:
LGTM

PR #124, opened 2016-10-18
Set h264Frame to null after we finish the frame

mjneil:
Currently, the `VideoSegmentStream` in the FLV transmuxer holds a reference to an `h264Frame` object across segment boundaries even though it always finishes the frame at the end of every flush. This means that on the next flush, that same frame will be finished a second time, causing a duplicate frame append. This also happens on replays where the last frame of the last segment will be added to the tags in the first segment.

This change sets `h264Frame = null` after every call to `finishFrame`. Also changed the metadata and extra data tags to have timestamps based on dts instead of pts to make sure tags are ordered correctly.

### DIFF
--- a/lib/flv/transmuxer.js
+++ b/lib/flv/transmuxer.js
@@ -196,13 +196,14 @@ VideoSegmentStream = function(track) {
     if (config && track && track.newMetadata &&
         (frame.keyFrame || tags.length === 0)) {
       // Push extra data on every IDR frame in case we did a stream change + seek
-      tags.push(metaDataTag(config, frame.pts));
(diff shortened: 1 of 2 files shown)


dmlap:
@mjneil is it hard to create a test case for this?

mjneil:
@dmlap I don't think it will be too hard. Mock two flushes, and check to make sure the second flush did not have the last frame from the first flush added to its list of tags. Cant check `h264Frame` itself since its inside the closure.

dmlap:
Ok, cool. Sounds like a good thing to add then.

mjneil:
Test added

forbesjo:
LGTM

PR #125, opened 2016-10-18
fix coalesce stream in FLV transmuxer to account for missing audio data in pending tracks

mjneil:
If there is an audio and video track in the segment but no audio data, (seen in several playlists with alternate audio tracks and no audio present in the main TS segments.), the `CoalesceStream` would never finish its flush process and emit a `done` event. This would cause the player to become stuck waiting to append to the buffer, with no errors. 

This change accounts for this possibility and allows `CoalesceStream` to complete its flush cycle with an empty audio track. I also pulled `CoalesceStream` out into its own file to help work towards needed project maintenance .

### DIFF
--- a/lib/flv/coalesce-stream.js
+++ b/lib/flv/coalesce-stream.js
@@ -1,0 +1,143 @@
+'use strict';
+
+var Stream = require('../utils/stream.js');
+
+/**
+ * The final stage of the transmuxer that emits the flv tags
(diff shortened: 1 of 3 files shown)


gesinger:
Looks good, though it would be good to have some unit tests.

forbesjo:
Looks good to me.

PR #130, opened 2016-11-30
Fill audio gaps at start of segment with silence

gesinger:
# Description

If an audio append start time is provided, and the next audio segment starts greater than a frame's duration from that time, but is within a half second from that time, fill the gap (up to the nearest whole frame) with silence.


jessetan:
When the `getSilentFrame` function was added to hls.js, the PR at https://github.com/dailymotion/hls.js/pull/412 noted that the frame payload that was used only differed with nr of channels, not with sampling rate. This was opposite to the code in mux.js at the time. There did not seem to be a conclusion on which was correct (also see comment https://github.com/videojs/mux.js/commit/f50426b257eb7d525311c1e6029e8891a2e07fe8#commitcomment-17302300)

imbcmdth:
LGTM

PR #133, opened 2016-12-08
Null captions

gesinger:


lib/m2ts/caption-stream.js
```diff
@@ -452,5 +460,6 @@ Cea608Stream.prototype.shiftRowsUp_ = function() {
 // exports
 module.exports = {
   CaptionStream: CaptionStream,
-  Cea608Stream: Cea608Stream
+  Cea608Stream: Cea608Stream,
+  parseCaptionPackets: parseCaptionPackets
```
mjneil: Why is this being exported now?

gesinger: Used for easier parsing of user data in other plugins (can skip CaptionStream)

gesinger: Actually, I can remove this.

PR #138, opened 2017-01-01
Remove GOP fusion. Fixes #136

### LINKED ISSUE #136: Chrome no longer requires fragments to start with a key frame
The Chrome bug was fixed back in February and shipped in release version 50. 
https://bugs.chromium.org/p/chromium/issues/detail?id=229412#c77

This code can now be removed (as well as code for other containers)
https://github.com/videojs/mux.js/blob/master/lib/mp4/transmuxer.js#L319

cc: @wolenetz

dmlap:
Chrome 50+ supports media segments that don't start with key frames. Remove the code that held onto old GOPs to prepend alongside new segments when they didn't begin with keyframes.


dmlap:
Tested this against a couple streams today and it doesn't look like we should pull this in. Chrome 50+ may accept media segments without leading keyframes but it doesn't seem like it's fusing GOPs across segment boundaries. I tried [advanced bip bop](https://d2zihajmogu5jn.cloudfront.net/bipbop-advanced/bipbop_16x9_variant.m3u8) using contrib-hls on Chrome55/Mac and saw numerous buffer gaps in video. The gap skipper managed to jump them but the experience is much worse than on the [demo page](http://videojs.github.io/videojs-contrib-hls/).

chcunningham:
Thanks for looking into it. I want to understand further to see if Chrome is behaving to-spec. Also, does the fusion occur automatically (without mux.js intervening) in other browsers?

> but it doesn't seem like it's fusing GOPs across segment boundaries

Can you describe the details of a GOP where fusion (not sure on the meaning) is expected but not occurring? For example, if a first segment with frames containing these presentation times (K = key, durations = 10 for all frames):
`0K 10 20 `

Then a second segment is appended with these times (note: no leading key frame):
`30 40 50`

I expect no gap will be created. Buffered range should report [0, 60).

dmlap:
@chcunningham I think Chrome is behaving to-spec. GOP fusion is a term we made up so no worries on that front. Basically, we retain the "leading edge" of the last GOP from the previous append and if we detect a segment begins with non-keyframes, we recreate a complete GOP from that info. So if we had segments that looked like this:

```
| Segment 1   | Segment 2
| 0...        | ...11...20

0...: GOP starting at time zero
11...: GOP starting at time 11
```

We'd "fuse" the GOP from 0 to 11 across the segment boundary. The appends that the SourceBuffer would see look like this:

```
| Append 1 | Append 2
| 0...     | 0...11...20
```

The resulting SourceBuffer would be buffered from 0 to 20. If we drop the fusion step, we'd have a gap between the end of Segment 1 and the keyframe at 11 seconds.

I believe FF and Edge handle this situation just like Chrome does now. It would be cool if MSE did this for us but the solution we have does seem to be working just fine.

chcunningham:
I think dropping the fusion code should be fine for this example. Chrome should place segment 2 right alongside segment 1 without creating any gaps. Before Chrome fixed its bug, this would definitely cause problems, but today this should work. Maybe I'm missing something...

dmlap:
Hmm. I'll play around with this a bit more then.

chcunningham:
Sure thing. If it helps, we have a [set of unit tests](https://cs.chromium.org/chromium/src/media/filters/chunk_demuxer_unittest.cc?q=f:chunk_demuxer_unit+RelaxedKeyframe&sq=package:chromium&dr=CSs&l=4428) that are pretty concise in documenting the behavior we expect (all with  prefix "RelaxedKeyframe").

Maybe you're hitting a bug we don't know about yet ;)

dmlap:
I removed code we had in place to trim [partial leading access units](https://github.com/videojs/mux.js/pull/138/files#diff-05211e7d4ced30d5a4c2514589989358R274) along with the GOP fusion stuff and tried it out on our debug page. The debug page does straight appends (into the SourceBuffer and it looked great there.

The debug page only appends a single init segment but contrib-hls appends one between each segment to workaround decode errors in Firefox when video parameters change. I'm guessing that's the critical difference here. We're going to look into detecting the relevant video changes in FF and only appending a new init segment in those cases. If that works, I think we can merge this in (yeah!).

chcunningham:
Woot! Yeah, I think segments following an init segment do need to begin with a key frame, so your reasoning makes sense to me.

PR #139, opened 2017-01-27
Update flv transmuxer for use in webworker

mjneil:
This changes the output of the flv transmuxer to be simple objects only containing the information videojs-contrib-media-sources needs to process and append the tags. Previously the transmuxer was just returning the full `FlvTag` objects, but you cannot send objects over a web worker that contain functions, so this strips those properties off. Also, `getFlvHeader` is now exported as a standalone function instead of being tied to an instance of the transmuxer. This way, contrib-media-sources can generate an flv header without an instance of the transmuxer since that instance will be living in a web worker.


lib/flv/index.js
```diff
@@ -1,4 +1,7 @@
+var transmuxer = require('./transmuxer');
+
 module.exports = {
   tag: require('./flv-tag'),
-  Transmuxer: require('./transmuxer')
+  Transmuxer: transmuxer.Transmuxer,
+  getFlvHeader: transmuxer.getFlvHeader
```
imbcmdth: Doing `transmuxer.Transmuxer` is confusing. I think it might be time to move `getFlvHeader` to it's own file. Then just import both of the files here.

imbcmdth:
LGTM!

lib/flv/flv-header.js
```diff
@@ -0,0 +1,60 @@
...
+
+var FlvTag = require('./flv-tag.js');
+
+// For information on the FLV format, see
+// http://download.macromedia.com/f4v/video_file_format_spec_v10_1.pdf.
+// Technically, this function returns the header and a metadata FLV tag
+// if duration is greater than zero
+// duration in seconds
+// @return {object} the bytes of the FLV header as a Uint8Array
+var getFlvHeader = function getFlvHeader(duration, audio, video) { // :ByteArray {
```
gesinger: Minor, shouldn't need to name the function.

lib/flv/tag-list.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+var TagList = function() {
```
gesinger: Does this need to have `length` too for `finishFrame`?

mjneil: good catch!

PR #140, opened 2017-02-11
Include raw packet in TransportParseStream data event

streamer45:
Adds a reference to the original raw TS packet in TransportParseStream data event object.
This is very useful in my case where I want to re-use the original packet based on the parsing information.


forbesjo:
Thanks for the PR but this goes outside the bound of what these streams are meant for. If you still need this feature we can look for an alternative for parsing this information.

PR #141, opened 2017-02-13
output extra flags on flv tags

mjneil:
some important information wasn't being attached to the tags outputted by the flv transmuxer. 
This adds them on

### DIFF
--- a/lib/flv/transmuxer.js
+++ b/lib/flv/transmuxer.js
@@ -197,8 +197,13 @@ VideoSegmentStream = function(track) {
     if (config && track && track.newMetadata &&
         (frame.keyFrame || tags.length === 0)) {
       // Push extra data on every IDR frame in case we did a stream change + seek
-      tags.push(metaDataTag(config, frame.dts).finalize());
-      tags.push(extraDataTag(track, frame.dts).finalize());
+      var metaTag = metaDataTag(config, frame.dts).finalize();
+      var extraTag = extraDataTag(track, frame.dts).finalize();
+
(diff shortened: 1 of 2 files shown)


lib/flv/transmuxer.js
```diff
@@ -197,8 +197,13 @@ VideoSegmentStream = function(track) {
     if (config && track && track.newMetadata &&
         (frame.keyFrame || tags.length === 0)) {
       // Push extra data on every IDR frame in case we did a stream change + seek
-      tags.push(metaDataTag(config, frame.dts).finalize());
-      tags.push(extraDataTag(track, frame.dts).finalize());
+      var metaTag = metaDataTag(config, frame.dts).finalize();
+      var extraTag = extraDataTag(track, frame.dts).finalize();
+
+      metaTag.metaDataTag = extraTag.metaDataTag = true;
```
gesinger: Why wasn't this needed before, but now is?

mjneil: Previously, the `FlashSourceBuffer` in contrib-media-sources would trim any flv tags that had a `pts` value that came before a calculated `targetPts` value. This was a non issue since these tags have the same `pts` value as the key frame they come before.  However, in https://github.com/videojs/videojs-contrib-media-sources/pull/124 the tag trimming logic was updated to only trim complete gops that came before the `targetPts` value. Since the `extraTag` here has `keyFrame = true` that caused the tag trimming logic to treat it as its own gop, causing these tags to get trimmed. This lets us know these are special meta data tags so we can account for them and not drop them if we are keeping the following key frame

imbcmdth:
LGTM

PR #142, opened 2017-02-15
Transmuxing can fail if there are less than two GOPs

imbcmdth:
`this.extendFirstKeyFrame_` should not throw an exception if there is only one GOP and should become a NOOP in that case.


mjneil:
LGTM

PR #143, opened 2017-02-28
Fix silence insertion to not insert extra frames when audio is offset…

gesinger:
… from video

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -218,16 +218,12 @@ AudioSegmentStream = function(track) {
     // determine frame clock duration based on sample rate, round up to avoid overfills
     frameDuration = Math.ceil(ONE_SECOND_IN_TS / (track.samplerate / 1024));
 
-    if (audioAppendStartTs &&
-        videoBaseMediaDecodeTime &&
-        // gap is at least a frame duration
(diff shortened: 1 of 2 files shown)


lib/mp4/transmuxer.js
```diff
@@ -215,14 +215,16 @@ AudioSegmentStream = function(track) {
...
 
     if (audioAppendStartTs &&
         videoBaseMediaDecodeTime &&
-        // old audio doesn't overlap with new audio
-        audioAppendStartTs < baseMediaDecodeTimeTs) {
-      audioGapDuration = baseMediaDecodeTimeTs - videoBaseMediaDecodeTime;
-      // determine frame clock duration based on sample rate, round up to avoid overfills
-      frameDuration = Math.ceil(ONE_SECOND_IN_TS / (track.samplerate / 1024));
+        // gap is at least a frame duration
+        baseMediaDecodeTimeTs - audioAppendStartTs >= frameDuration) {
```
imbcmdth: I feel like this condition should be:

`baseMediaDecodeTimeTs - Math.max(audioAppendStartTs, videoBaseMediaDecodeTime) >= frameDuration` 

...but if that is the case then we can just remove the condition completely since it'll be calculated below and, if this were to evaluate to false, would result in a value that is `audioFillFrameCount < 1` and we will exit early.

forbesjo:
Looks good to me

PR #146, opened 2017-03-29
reset TimestampRolloverStream on discontinuities

mjneil:
https://github.com/videojs/videojs-contrib-hls/issues/1056

The above issue stems from two key things. 

 * the `TimestampRolloverStream` doesn't reset its `referenceDTS` value on discontinuities. This wasn't really an issue because of the way we use `baseMediaDecodeTime` to shift segments to the correct time on discontinuities, so it didn't matter if the transmuxer treated a segment on a discontinuity as a rollover since `baseMediaDecodeTime` would shift it correctly regardless.

 * each `type` (`video`, `audio`, `timed-metadata`) has its own independent `referenceDTS`. This is important because the rollover for video may be in a different segment than the rollover in audio if it is near segment boundaries. However, in the case of the issue linked above, because only the ads have ID3 tags, and the main content doesn't, and because `referenceDTS` wasn't reset on discontinuities, the `videoTimestampRolloverStream` was detecting a rollover in video, but the `timedMetadataTimestampRolloverStream` never had `referenceDTS` set, so it did not detect a rollover, giving the ID3 tags a different PTS value than the video. When `baseMediaDecodeTime` was being used to shift the times later on in the pipeline, the ID3 tags were being shifted too much into the negative values.

By resetting all the `TimestampRolloverStream`s on discontinuities, we can avoid this edge case

### DIFF
--- a/lib/m2ts/timestamp-rollover-stream.js
+++ b/lib/m2ts/timestamp-rollover-stream.js
@@ -69,6 +69,11 @@ var TimestampRolloverStream = function(type) {
     this.trigger('done');
   };
 
+  this.discontinuity = function() {
+    referenceDTS = void 0;
+    lastDTS = void 0;
+  };
+
 };
 
 TimestampRolloverStream.prototype = new Stream();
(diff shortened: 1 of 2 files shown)


zhuangs:
Tested and LGTM.

PR #148, opened 2017-04-04
Update metadata-stream.js so TXXX are not clipped

RevinKey:
## Description
Parsing of TXXX tag is cutting off last character of tag. 
Do not need to account for a null terminator.

## Sample with Brightcove (videojs)player:
current: http://cs1.brightcodes.net/kreynolds/BrightcoveExamples/test/414473/current.html
proposed: http://cs1.brightcodes.net/kreynolds/BrightcoveExamples/test/414473/index.html

## Specific Changes proposed
change line48 in file metadata-stream.js to 
          tag.value = parseUtf8(tag.data, i + 1, tag.data.length).replace('\0','');

## Requirements Checklist
- [ ] Feature implemented 
- [ ] If necessary, more likely in a feature request than a bug fix
  - [x] Example created ([starter template on JSBin](http://jsbin.com/axedog/edit?html,output))

### DIFF
--- a/lib/m2ts/metadata-stream.js
+++ b/lib/m2ts/metadata-stream.js
@@ -44,8 +44,6 @@ var
                 if (tag.data[i] === 0) {
                     // parse the text fields
                     tag.description = parseUtf8(tag.data, 1, i);
-
-
                     // do not include the null terminator in the tag value
                     tag.value = parseUtf8(tag.data, i + 1, tag.data.length).replace(/\0*$/, '');
                     break;
--- a/lib/m2ts/metadata-stream.js
+++ b/lib/m2ts/metadata-stream.js
@@ -47,7 +47,7 @@ var
 
 
           //do not include the null terminator in the tag value
-          tag.value = parseUtf8(tag.data, i + 1, tag.data.length).replace(/\0*$/,'');
+          tag.value = parseUtf8(tag.data, i + 1, tag.data.length).replace(/\0*$/, '');
           break;
         }
       }
--- a/lib/m2ts/metadata-stream.js
+++ b/lib/m2ts/metadata-stream.js
@@ -44,8 +44,10 @@ var
         if (tag.data[i] === 0) {
           // parse the text fields
           tag.description = parseUtf8(tag.data, 1, i);
-          // do not include the null terminator in the tag value
-          tag.value = parseUtf8(tag.data, i + 1, tag.data.length - 1);
+
+
+          //do not include the null terminator in the tag value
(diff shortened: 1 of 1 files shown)


imbcmdth:
The null-terminator **may** be present according to the spec. There are instances where it should be ignored (such as in the test cases that are failing). The code should check for and remove any null-termination before processing the string.

RevinKey:
Would this be enough for parsing out the null character:
 tag.value = parseUtf8(tag.data, i + 1, tag.data.length).replace('\0','');

In my test it works. I updated my player script in my sample to reflect this change as well if you would like to view it in action.

imbcmdth:
Thanks!

Looks like it works but I worry about multiple trailing nulls. That replace will only replace the first instance of `\0` it finds in the string. I think a better solution might be `.replace(/\0*$/, '');` to replace *one or more* null characters that appear at the **end** of the string.

RevinKey:
Cool. My Branch is updated and passed Travis. Let me know if you need anything else.

lib/m2ts/metadata-stream.js
```diff
@@ -5,243 +5,246 @@
...
+                return;
+            }
+
+            for (i = 1; i < tag.data.length; i++) {
+                if (tag.data[i] === 0) {
+                    // parse the text fields
+                    tag.description = parseUtf8(tag.data, 1, i);
+
+
+                    //do not include the null terminator in the tag value
```
imbcmdth: This is failing linting in travis because it's missing a space after the `//`.

imbcmdth:
@RevinKey Seems like the indenting on the file changed and there are a few extra newlines (surprised that the linting didn't complain about either!)

Compare: https://github.com/videojs/mux.js/pull/148/files
With: https://github.com/videojs/mux.js/pull/148/files?w=1

And you can see how many lines had nothing changed except newlines. Would you mind getting this PR back down to 2 or three lines changed?

lib/m2ts/metadata-stream.js
```diff
@@ -5,243 +5,246 @@
...
+                // ignore frames with unrecognized character encodings
+                return;
+            }
+
+            for (i = 1; i < tag.data.length; i++) {
+                if (tag.data[i] === 0) {
+                    // parse the text fields
+                    tag.description = parseUtf8(tag.data, 1, i);
+
+
```
imbcmdth: Extra newline.

imbcmdth:
@RevinKey Could you add a test for the non-null terminating case? It should be pretty simple to take an existing test and creating a new one from that. As for all the whitespace changes, it might be easier to just start over and do the change again.

PR #150, opened 2017-04-11
CEA608: support for cc2-cc4, special/extended characters, formatting

squarebracket:
I know some JavaScript, but I'm new to node, so it's possible I'm not obeying proper code style in some places. Please do let me know if that's the case. Also, maybe relax those lint rules since there were over 400 warnings when I cloned the project....

Some of the added features introduced a bit of duplication into the code, but I think this is reasonable as it makes the overall code paths easier to understand.

### Added Features

1. All four CC tracks are now available

There are now four Cea608Stream objects attached to the caption
stream. Each one has some setup code that sets the bytes used
for parsing out track information. Each Cea608Stream receives
every caption packet, but the Cea608Stream will `return` as soon
as possible so that it doesn't spend time processing packets that
target another track.

The active channel is tracked with a global variable, which is set
when a Cea608Stream receives a control code destined for its channel.

2. Special and Extended Characters

Before, the Cea608Stream would consider all special and extended
characters (as defined in CEA608) as unsupported control codes. It will
not properly translate those characters to their unicode equivalent.

3. Formatting

Underline and italics will now be rendered. These are tracked internally
by

Additionally, PAC rows are now properly interpreted, and text is
inserted into the correct row in the buffer. Pop-on captions should now
render line breaks at the correct spot. Instead of removing _all_ blank
lines, only those at the start or end of the caption are removed.

### Misc Code Stuff

You may notice that I have some code paths that do essentially nothing. Though some cases could probably be collapsed into a single "unsupported" case, I think it's best to keep things as is in case someone wants to extend support in the future.

Also, I've added a `track` property to the caption object that's pushed. This is so that https://github.com/videojs/videojs-contrib-media-sources can tell which captions are meant for which track. I have added that functionality to `media-sources` and will submit a PR there too.

### Still Unsupported
1. Colors
But seriously who does that.

2. Custom Whitespacing
I had implemented interpreting of the PACs and control codes for indentation, but realized that it was mostly just used for center alignment. I'm not sure of how to differentiate between that, and captions that legitimately use whitespace to improve comprehension -- like captions on the left and right side of the screen for the characters that are in those positions. If anyone has any ideas on such logic, I might see if I can implement it. For now, I just left it as is.

3. Paint-on Mode


squarebracket:
Just found a bug in roll-up introduced during a final clean-up that the unit tests don't catch, don't merge yet.

squarebracket:
The bug is happening because of a caption that has `startPts == endPts`. I think that such captions should just be dropped in roll-up mode, as they're the result of very short lines.

squarebracket:
Note: see https://github.com/videojs/videojs-contrib-media-sources/pull/140 for related functionality in videojs-contrib-media-sources.

lib/m2ts/caption-stream.js
```diff
@@ -402,47 +662,67 @@ Cea608Stream.prototype.flushDisplayed = function(pts) {
...
 };
 
+// Adds the opening HTML tag for the passed character to the caption text,
+// and keeps track of it for later closing
+Cea608Stream.prototype.addFormatting = function(pts, format) {
+  var text = '';
+  for (var i = 0; i < format.length; i++) {
+    text += '<' + format[i] + '>';
+  }
+  this.formatting_ += format;
```
imbcmdth: If `this.formatting_` was always an array of single characters it would make things easier later when you want to unset them because you can use the much prettier iterator functions!

```javascript
Cea608Stream.prototype.addFormatting = function(pts, format) {
  this.formatting_ = this.formatting_.concat(format);
  
  return format.reduce(function (text, format) {
    return text + '<' + format + '>';
  }, '');
};

Cea608Stream.prototype.clearFormatting = function(pts) {
  if (!this.formatting_.length) {
    return;
  }
  const text = this.formatting_.reverse().reduce(function (text, format) {
    return text + '</' + format + '>';
  }, '');
  this.formatting_ = [];
  this[this.mode_](pts, text);
};
```

test/caption-stream.test.js
```diff
@@ -798,7 +1088,167 @@ QUnit.test('preamble address codes are converted into spaces', function() {
   ].forEach(cea608Stream.push, cea608Stream);
 
   QUnit.equal(captions.length, 1, 'caption emitted');
-  QUnit.equal(captions[0].text, '01 02', 'PACs were converted to space');
+  QUnit.equal(captions[0].text, '0102', 'PACs were NOT converted to space');
+});
+
+QUnit.test('preserves newlines from PACs in pop-on mode', function() {
```
imbcmdth: WOW! Amazing tests!

test/caption-stream.test.js
```diff
@@ -282,12 +282,92 @@ QUnit.test('converts non-standard character codes to ASCII', function() {
   });
 
   packets.forEach(cea608Stream.push, cea608Stream);
-
   QUnit.equal(captions[0].text,
         String.fromCharCode(0xe1, 0xe9, 0xed, 0xf3, 0xfa, 0xe7, 0xf7, 0xd1, 0xf1, 0x2588),
         'translated non-standard characters');
 });
 
+QUnit.test('converts special character codes to ASCII', function() {
```
imbcmdth: Awesome!

lib/m2ts/caption-stream.js
```diff
@@ -264,9 +370,56 @@ var createDisplayBuffer = function() {
...
+  //           chars (ext_+0, ext_+1, and ext_+2 are char0s for
+  //           extended codes)
+  // control_: char0 for control codes, except byte-shifted to the
+  //           left so that we can do this.control_ | CONTROL_CODE
+  // offset_:  char0 for tab offset codes
+  //
+  // It's also worth noting that I've found control codes for field-2
+  // captions to be spec-breaking; they should be 0x14 or 0x1c, but
+  // come as 0x15 and 0x1d. That's why there's the "| field" on the
+  // control value.
```
squarebracket: Woops! Thanks for catching this. I'd written this comment mostly as a note to myself so that I didn't forget to mention it in the PR. I later found where in the CEA608 spec it mentions this, but neglected to do update this comment with that information, which I will do now.

lib/m2ts/caption-stream.js
```diff
@@ -264,9 +370,56 @@ var createDisplayBuffer = function() {
   return result;
 };
 
-var Cea608Stream = function() {
+var Cea608Stream = function(track) {
   Cea608Stream.prototype.init.call(this);
 
+  this.track_ = track || 1;
+  switch (this.track_) {
```
squarebracket: I actually originally passed in `field` and `dataChannel`, but IIRC I changed it for a couple reasons. One was just that it looked nicer when constructing the objects (`this.ccn_ = new Cea608Stream(n);`), but more importantly, it's passed along to videojs-contrib-media-sources:

```javascript
this.trigger('data', {
  startPts: this.startPts_,
  endPts: pts,
  text: content,
  track: this.track_
});
```

I think that makes way more sense than passing the field and data channel. That being said, [videojs-contrib-media-sources is just using it as the track name](https://github.com/videojs/videojs-contrib-media-sources/pull/140/files#diff-cf1b209dfcf83384fb52b6a3c0eb5047R23). So maybe we could do something like this:

```javascript
var this.cc1_ = new Cea608Stream(0, 0);
...
var Cea608Stream = function(field, dataChannel) {
  // this next line is ugly, but should produce the correct track #
  this.name_ = 'cc' + (((field << 1) | dataChannel) + 1);
  this.field_ = field;
  this.dataChannel_ = dataChannel;
...
}
Cea608Stream.prototype.flushDisplayed = function(pts) {
...
  this.trigger('data', {
    startPts: this.startPts_,
    endPts: pts,
    text: content,
    track: this.name_
  });
}
```

Thoughts?

lib/m2ts/caption-stream.js
```diff
@@ -319,77 +494,162 @@ var Cea608Stream = function() {
...
+              this.clearFormatting(packet.pts);
+              this.row_ = index;
+            }
+
+            if ((data & 0x10) === 0x10) {
+              // we've got an indent level
+              this.column_ = ((data & 0xe) >> 1) * 4;
+            }
+
+            return; // we've found the row so break out of this function
```
squarebracket: I guess you mean due to your above-mentioned suggestion? Because, unless JavaScript has somehow gained omniscient powers, I don't know how it would know to break out of the for loop before it has gone through all items?

lib/m2ts/caption-stream.js
```diff
@@ -319,77 +494,162 @@ var Cea608Stream = function() {
...
-      }
-
       // character handling is dependent on the current mode
-      this[this.mode_](packet.pts, char0, char1);
-      break;
+      // note that we only get here when we'll actually be adding text
+      // to the caption
+      this[this.mode_](packet.pts, text);
+
+      break; // ends `default` switch block
```
squarebracket: I entirely agree that the `default` case is unwieldy. Honestly, I think the switch should go in favour of if/else blocks. I was hesitant to remove it only because, like you say, it was like that when I got here.

A refactor to if/else would, I think, make a much more logical code flow, because all "handle X" logic would be at the top level. Something like this:

```javascript
var char0 = data >> 8;
var char1 = data & 0x00ff;
if (data == this.RU2) {
  // do roll-up things
} else if (data == this.RCL) {
  // do pop-on things
} else if (char0 == this.specialCharacters && char1 >= 0x30 && char1 <= 0x3f) {
  // append special character
} else if (char0 === this.ext_ && (char1 >= 0x20 && char1 <= 0x2f)) {
  // handle mid-row codes
} else {
  // we've got an unsupported code, so just return
  return;
}
```

lib/m2ts/caption-stream.js
squarebracket: Dat JavaScript magic. Thanks, those are great suggestions.

lib/m2ts/caption-stream.js
```diff
@@ -319,77 +494,162 @@ var Cea608Stream = function() {
...
+        // Bitmask char0 so that we can apply character transformations
+        // regardless of field and data channel.
+        // Then byte-shift to the left and OR with char1 so we can pass the
+        // entire character code to `getCharFromCode`.
+        char0 = (char0 & 0x03) << 8;
+        text = getCharFromCode(char0 | char1);
+        this.column_++;
+
+      // Handle extended characters
+      } else if ((char0 === (this.ext_ + 1) || char0 === (this.ext_ + 2)) &&
```
squarebracket: Hmmmmm.......... While I agree with you that `char0 === (this.ext_ + 1)` isn't easy to read or understand, I don't particularly like `this.extSetSpanishFrenchMisc_ = 0x13` either. Someone unfamiliar with the spec might mistakenly understand that to mean that `0x12` indicates an extended character, which is not the case -- it can also be a PAC. This is even more confusing for `0x11`, which can be a special character, a mid-row code, or a PAC.

If you think the comments aren't enough, how do you feel about object methods? Something like
```javascript
if (this.isSpecialCharacter(data)) {
  // do the thing
} else if (this.isExtCharacter(data)) {
  // do the other thing
} // etc...

/**
 * Detects if the 2-byte packet data is a special character
 *
 * Special characters have a second byte in the range 0x30 to 0x3f,
 * with the first byte being 0x11 (for data channel 1) or 0x19 (for
 * data channel 2).
 *
 * @param  {Integer} data  The two cc_data bytes
 * @return {Boolean}       Whether the data is an special character
 */
Cea608Stream.prototype.isSpecialCharacter = function(char0, char1) {
  return (char0 === this.ext_ && char1 >= 0x30 && char1 <= 0x3f);
}

/**
 * Detects if the 2-byte packet data is an extended character
 *
 * Extended characters have a second byte in the range 0x20 to 0x3f,
 * with the first byte being 0x12 or 0x13 (for data channel 1) or
 * 0x1a or 0x1b (for data channel 2).
 *
 * @param  {Integer} data  The two cc_data bytes
 * @return {Boolean}       Whether the data is an extended character
 */
Cea608Stream.prototype.isExtCharacter = function(char0, char1) {
  return ((char0 === (this.ext_ + 1) || char0 === (this.ext_ + 2)) &&
    (char1 >= 0x20 && char1 <= 0x3f));
}
```

That way the general flow is nice and abstracted, but the gory implementation bits are in the corresponding function.

lib/m2ts/caption-stream.js
squarebracket: I actually convinced myself that this is the best option.

squarebracket:
I made many of the changes you requested, except for the couple of ones I commented on above.

I also want to implement a test to make sure that data received before a control code (that sets the active channel) are dropped. The problem is that, since `activeChannel` is a global variable, it's set earlier in the testing file. Is there any way around that?

lib/m2ts/caption-stream.js
imbcmdth: I agree with this! 👍

lib/m2ts/caption-stream.js
imbcmdth: I had to look it up myself! It's been a while since I 608'd and wanted to understand what the heck was happening here.

lib/m2ts/caption-stream.js
imbcmdth: I do think this is the better solution!

lib/m2ts/caption-stream.js
imbcmdth: The function supplied to `forEach` is called once per item in the array. There is no way to early exit in JavaScript.

lib/m2ts/caption-stream.js
imbcmdth: Agreed! This is a good solution.

lib/m2ts/caption-stream.js
```diff
@@ -139,17 +140,29 @@ var parseCaptionPackets = function(pts, userData) {
   return results;
 };
 
+var activeChannel = null;
```
imbcmdth: After reading your comment about the difficulty testing this because it's (module) global, I realized that this is going to have to change.

The goal is to have some state that is shared between all 4 `Cea608Stream` objects. Here is what I am thinking: 
* Make the `activeChannel` state a property of the `CaptionStream`
* Have the `Cea608Stream` objects emit an event like "activechannelchange" that alerts the `CaptionStream` object that it needs to change the `activeChannel`
* Have the `CaptionStream` call each `Cea608Stream#setActiveChannel` to update each stream with the new active channel information.

lib/m2ts/caption-stream.js
```diff
@@ -319,77 +440,157 @@ var Cea608Stream = function() {
       this.startPts_ = packet.pts;
       break;
 
-    case ROLL_UP_2_ROWS:
+    case (this.ROLL_UP_2_ROWS_):
```
imbcmdth: It isn't necessary to have the case wrapped in parentheses.

squarebracket:
Hmm. Yeah, that sounds like a reasonable solution. The asynchronous call should be done processing for all Cea608Streams by the time the next packet comes in, right?

imbcmdth:
There are no real asynchronous calls here. All processing is done synchronously and the "stream" and event abstractions are just a way to connect pieces in a consistent manner.

squarebracket:
Cool, I'll wire things up like that then, along with the rest of the changes.

imbcmdth:
Sweet! 🍨

squarebracket:
I can't think of a way of having a `Cea608Stream` having reference to the `CaptionStream` that contains it, other than passing `this` to its constructor. Is there a JavaScript-ier way of doing it that I don't know about?

Another simpler approach might just be
```javascript
if (this.setsThisDataChannelActive(data)) {
  this.active_ = true;
} else if (this.setsThisDataChannelInactive(data)) {
  this.active_ = false;
  return; // we shouldn't process the data
} else if (!this.active_) {
  return;
}
```

Thoughts?

squarebracket:
Here is my proposed solution the `activeChannel` thing: https://github.com/squarebracket/mux.js/compare/cea608-improvements...squarebracket:activechange-captioncontroller

squarebracket:
After having it kicking around in my head for a while, I decided to do a bit of refactoring. All stream selection logic is now inside of `CaptionStream` -- each packet is pushed to only one `Cea608Stream`. This is done by simply calling `this.activeCea608Streams_[packet.type].push(packet)`. Since we need to drop all packets before receiving a control code that sets the active channel, `activeCea608Streams` is initialized to contain two `push` functions that do nothing. Once that control code is received, `activeCea608Streams` contains either CC1, CC3 or CC2, CC4.

The refactoring also makes some variable names clearer as being specific to 608, and it should make integrating a 708 parser into this code a painless process.

Because of the logic move, I had to move the "drops packets before first control code" and "ignore 708 packets" tests up into the `CaptionStream` area of the tests.

If everything looks good to you, I'll squash into a single commit.

squarebracket:
Unfortunately, this has been broken by video.js >= 6.2.0. When it comes across cues with formatting, it throws an error to the console:
```
VIDEOJS: ERROR: TypeError: Cannot assign to read only property 'localName' of object '#<HTMLElement>'
    at createElement (video.js:8292)
    at parseContent (video.js:8331)
    at new CueStyleBox (video.js:8509)
    at video.js:8968
    at Function.WebVTT$1.processCues (video.js:8980)
    at TextTrackDisplay.updateForTrack (video.js:12017)
    at TextTrackDisplay.updateDisplay (video.js:11988)
    at Player.bound (video.js:2203)
    at HTMLDivElement.bound (video.js:2203)
    at HTMLDivElement.data.dispatcher (video.js:1843)
```

Captions without formatting seem to work fine.

mjneil:
I am unable to reproduce the console errors related to formatting

squarebracket:
Interesting.... I have tried compiling 6.2.5 and the CDN's 6.2.4 and I get the error. So you see italics rendered?

test/caption-stream.test.js
```diff
@@ -242,6 +245,104 @@ QUnit.test('sorting is fun', function() {
...
+      captions.push(caption);
+    });
+  });
+
+  var seiNals = multichannelCaptions.map(makeSeiFromCaptionPacket);
+  seiNals.forEach(captionStream.push, captionStream);
+  captionStream.flush();
+
+  QUnit.equal(captions.length, 3, 'parsed three captions');
+  QUnit.equal(captions[0].text, 'BUT IT\'S NOT SUFFERING\nRIGHW.', 'parsed first caption correctly');
```
mjneil: is this caption intentionally misspelling `RIGHT` as `RIGHW`?

test/caption-stream.test.js
```diff
@@ -514,6 +701,197 @@ QUnit.skip('applies preamble address codes', function() {
...
+    { ccData: characters('ye'), type: 0 },
+    { ccData: characters('s.'), type: 0 },
+    { ccData: 0x142d, type: 0 }
+  ];
+
+  packets.forEach(cea608Stream.push, cea608Stream);
+  QUnit.equal(captions[0].text, 'no <i><u>yes.</u></i>', 'properly closed by CR');
+});
+
+QUnit.test('applies PAC color underline', function() {
```
mjneil: Bit confused here. Is the color portion of this (and the following 2 tests) being tested asserted here? Your PR description mentions that color is not implemented yet. Are these tests because only white text is supported?

test/caption-stream.test.js
squarebracket: Yup. It was like that when I got here :innocent:

It comes from an external file and I didn't feel like reprocessing the caption by hand.

mjneil: I see this was just taken from the bottom of this file and moved up here, so I assume yes as this test has always been like this

test/caption-stream.test.js
squarebracket: Good point about the ambiguous name. I say "PAC color underline" because coloured underline PACs are processed, and underline is added to the text. However, as you point out, colour is not supported, so that information is dropped.

So if you were to change the PAC from `0x1461` (row 15 white underline) to `0x1463` (row 15 green underline), you will still get out `'<u>yes.</u>'`. If colour support is added, more tests would be needed to differentiate between the various supported colours.

mjneil: Makes sense, would be helpful to just add a comment mentioning that the colored PACs are processed, but the color specific information is dropped

squarebracket: Actually, after going back over things, I renamed it to "applies PAC underline" since it is applied for both colour PACs and indentation PACs.

squarebracket:
Pushed a couple tiny commits to address the comments above.

mjneil:
You could probably make use of this test https://github.com/videojs/mux.js/blob/master/test/caption-stream.test.js#L180
check `data.captionTracks` The data needs to go through the whole transmuxer since this change was made in the `CoalesceStream`

lib/mp4/transmuxer.js
```diff
@@ -976,6 +978,7 @@ CoalesceStream.prototype.flush = function(flushSource) {
     caption.startTime /= 90e3;
     caption.endTime = (caption.endPts - timelineStartPts);
     caption.endTime /= 90e3;
+    event.captionTracks[caption.track] = true;
```
mjneil: You will have to make these changes in the `CoalesceStream` for the FLV transmuxer as well https://github.com/videojs/mux.js/blob/master/lib/flv/coalesce-stream.js#L112

mjneil: I believe this should also be `caption.stream` `caption.track` is `undefined`

squarebracket:
Good idea for the test location. I'm not sure how I'm going to get the caption data into an appropriate format, though....

mjneil:
Hmm, I was just thinking to use the `sintelCaptions` already being used. Not sure how that file was originally created

squarebracket:
I could, but ideally, I'd like to use something that has more than 1 caption track.

mjneil:
Ya thats probably for the best. Maybe you could use ffmpeg to segment your segment into smaller chunks to get a small amount of data that has captions. Or since ts segments are in a streaming format, you could maybe just trim off the end of the file, saving only the first x bytes, enough to have the caption data you want.

squarebracket:
I'll dig back into the 708 spec and modify a few captions by hand in the sintel file... that should be good enough I think.

mjneil:
I've noticed that rendition switches cause duplicated captions for the period of time that was buffered prior to the rendition switch. I think this is because the buffer is not cleared on a switch, and since we process segments on the new quality we've already processed on the previous so that visual switch happens faster, the captions already parsed and added to the track are still in the track. e.g. buffer 30 seconds on quality, upswitch, start processing new quality, for the next 30 seconds you will see double captions until you exhaust the 30 seconds worth of captions already added.

Though this is slightly different problem than #158 is describing since they aren't necessarily across segments and still being rendered in proper order, but I think it will help solve this issue as well.

squarebracket:
That condition you're describing was the main motivation for #158,
actually. Other conditions that cause that behaviour are more rare.

Do you want me to finish up the unit tests on that PR so it can be staged
first?

I originally created #158 on top of this PR and then reworked it for
upstream, so either way...

squarebracket:
Oh... just saw your edits

mjneil:
You can leave #158 on top of this as order is not too important, but I would like to get both PRs in before versioning and publishing a release. This PR is looking good to me

squarebracket:
Cool, I will squash and edit the commit message for the changes made since 1st commit.

squarebracket:
Actually, do you think I should rename `captionTracks` to `captionStreams` to be consistent with the `caption.stream` property?

mjneil:
I do think it would be good to have a single terminology that is used across the board. Based on the wording of the document you referenced in code comments https://www.gpo.gov/fdsys/pkg/CFR-2007-title47-vol1/pdf/CFR-2007-title47-vol1-sec15-119.pdf it sounds like the term `channel` is used. I would be open to using any of `caption.channel`, `caption.stream`, or `caption.track`, and the `event.captionChannel` (insert stream or track respectively). You probably have a better knowledge of CEA608 for which term would be most appropriate

squarebracket:
In the specs, there doesn't seem to be a single word they use. I mean, it's a hardware spec, so they probably just see `CC1` -`CC4` as the boxes in a K-map for `dataChannel` and `field`.

I'm gonna go with `stream` since:
- there's no confusion with `dataChannel`
- `track` is used plenty
- the HLS spec uses `INSTREAM-ID`
- it's analogous to a video/audio PES

PR #153, opened 2017-05-11
ts probe searches packets for first it can successfully parse

mjneil:
Sometimes audio or video packets can be padded with filler bytes to maintain a specific bitrate. When the TS Probe is inspecting packets to determine the audio/video end time of a segment, it starts with the last packet and grabs the timestamp of the first audio/video packet it sees. If the segment has been padded with filler bytes, the packet may not contain a timestamp causing NaN issues. 

This change adds a return of `null` during parsing a PES packet's timestamps when the offset length indicated by the adaption field is larger than length of the packet.


lib/m2ts/probe.js
```diff
@@ -121,6 +121,11 @@ var parsePesTime = function(packet) {
 
   var offset = 4 + parseAdaptionField(packet);
 
+  if (offset >= packet.byteLength) {
```
gesinger: Although the note below is a succinct description, it might be useful to expand a bit and say that if a packet contains data, it will have pts/dts values. If a packet doesn't have data, and is just stuffing bytes (0xFF), then the adaptation field length will expand to fill the rest of the packet, therefore the offset will reach the end of the packet.

lib/tools/ts-inspector.js
```diff
@@ -105,9 +105,11 @@ var parseAudioPes_ = function(bytes, pmt, result) {
           pusi = probe.ts.parsePayloadUnitStartIndicator(packet);
           if (pesType === 'audio' && pusi) {
             parsed = probe.ts.parsePesTime(packet);
-            parsed.type = 'audio';
-            result.audio.push(parsed);
-            endLoop = true;
+            if (parsed !== null) {
```
gesinger: Probably fine to just check for a truthy value (and below).

Issue #155, opened 2017-06-14
Fixed-point calculation of tkhd's fractional height and width is wrong.


mdesjardins:
I'm using version 4.1.3 of mux.js. I ran into a problem today where the height and width in the tkhd boxes were being calculated incorrectly by the MP4 Inspector. It turns out the fractional portion of the tkhd box's width and height were being divided by 16 instead of 2^16 (65536) - perhaps it was intended to be a shift right?  Anyhow, I have a PR with a fix that I can submit, I'll do so in a second.

PR #156, opened 2017-06-14
Fixes fixed point calculation of tkhd box's height and width.

mdesjardins:
This is a fix for issue videojs/mux.js#155 ... the fractional portion of the tkhd box's height and width was being calculated incorrectly (divided by 16, instead of shifting right by 16 or divided by 65536).

### DIFF
--- a/lib/tools/mp4-inspector.js
+++ b/lib/tools/mp4-inspector.js
@@ -594,9 +594,9 @@ var
       i += 2;
       result.matrix = new Uint32Array(data.subarray(i, i + (9 * 4)));
       i += 9 * 4;
-      result.width = view.getUint16(i) + (view.getUint16(i + 2) / 16);
+      result.width = view.getUint16(i) + (view.getUint16(i + 2) / 65536);
       i += 4;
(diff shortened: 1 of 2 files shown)


imbcmdth:
Thanks a lot for the PR. It looks good!

Out of curiosity, what was the issue you were running into with this? Since we don't use this file directly in our transmuxer, I suspect you were doing something very cool to have stumbled across this! 👍

mdesjardins:
@imbcmdth I'm actually using the inspector to validate, client-side, that a video uses acceptable parameters for a service to which I'm uploading the video. :)

imbcmdth:
@mdesjardins Very awesome idea! We will try and get this merged soon. Thanks again.

forbesjo:
Sorry we lost track of this, we'll take a look at this soon, add another test and merge in your change.

forbesjo:
Do you have happen to have content to confirm the fix?

gkatsev:
@mdesjardins hey, do you happen to have test content for this?

@imbcmdth do you have any ideas how we can test this so it can finally be merged?

Issue #105, opened 2016-08-01
Issues with streams that have PES packets split between segments


ntadej:
Our streams have audio PES packets split between segments - one starts at the end of one .ts file and continues into the next. Calling flush on audio in ElementaryStream (https://github.com/ntadej/mux.js/blob/master/lib/m2ts/m2ts.js#L434) causes player to stall for a brief moment (throbber is shown). Commenting this line out fixes all playback issues with our streams that I reported with new videojs-contrib-hls. (it breaks seeking though)

I'm not familiar enough with mux.js to fix this on my own but I'm happy to provide you any assistance needed to do so. You should already have access to our steams.

mjneil:
Sorry for the long time to resolve this, but this should be fixed by #157 which is now released

PR #157, opened 2017-06-23
Only flush PES packets from TS parsing front end when they are complete

imbcmdth:
Where `complete` is defined as any time `PES_packet_length` matches the data’s length OR is a video packet.

Works around an issue with incomplete packets getting sent down the pipeline when the source has audio PES packets split between segments.

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -321,7 +321,7 @@ ElementaryStream = function() {
         event = {
           type: type
         },
-        i = stream.data.length,
+        i = 0,
         offset = 0,
         packetFlushable = false,
         fragment;
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -270,6 +270,9 @@ ElementaryStream = function() {
     parsePes = function(payload, pes) {
       var ptsDtsFlags;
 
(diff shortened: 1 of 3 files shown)


imbcmdth:
This is also a fix for #105

lib/m2ts/m2ts.js
```diff
@@ -312,35 +315,50 @@ ElementaryStream = function() {
...
     },
-    flushStream = function(stream, type) {
+    flushStream = function(stream, type, forceFlush) {
       var
         packetData = new Uint8Array(stream.size),
         event = {
           type: type
         },
-        i = 0,
+        i = stream.data.length,
```
imbcmdth: ARGG!  let me fix this...

lib/m2ts/m2ts.js
```diff
@@ -312,35 +315,50 @@ ElementaryStream = function() {
...
+        packetData.set(fragment.data, offset);
+        offset += fragment.data.byteLength;
       }
 
       // parse assembled packet's PES header
       parsePes(packetData, event);
 
-      stream.size = 0;
+      // non-video PES packets MUST have a non-zero PES_packet_length
+      // check that tthey match before we do a flush
```
dmlap: `s/tthey/they/`

PR #158, opened 2017-07-21
Caption fixes (incl duplicate caption processing)

squarebracket:
Sometimes the same segment is pushed twice to the transmuxer. For instance, an initial seek pushes the first segment twice, and a rate switch sometimes pushes the same segment but at two different rates. As you might expect, this messes with caption processing.

The general idea is that we track the most recently received DTS, and if we receive any DTS less than that, we return without processing the caption data. There is a caveat, which is when data for single DTS is split across segments. For that case, we need to ignore the first packet, but not anything after, since it is a new segment. See the comments in the code.

I've marked this as WIP since I'm not sure if my assumption about monotonically increasing DTS values is correct. Also, this breaks when seeking back in time, and I'm not sure how to fix that.

### DIFF
--- a/lib/flv/transmuxer.js
+++ b/lib/flv/transmuxer.js
@@ -400,6 +400,11 @@ Transmuxer = function(options) {
     packetStream.flush();
   };
 
+  // Caption data has to be reset when seeking outside buffered range
+  this.resetCaptions = function() {
+    captionStream.reset();
+  };
+
   // Re-emit any data coming from the coalesce stream to the outside world
   coalesceStream.on('data', function(event) {
     self.trigger('data', event);
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -1213,6 +1213,7 @@ Transmuxer = function(options) {
       videoTrack.timelineStartInfo.dts = undefined;
       videoTrack.timelineStartInfo.pts = undefined;
       clearDtsInfo(videoTrack);
+      pipeline.captionStream.reset();
       videoTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
     }
 
@@ -1248,6 +1249,14 @@ Transmuxer = function(options) {
     // Start at the top of the pipeline and flush all pending work
     this.transmuxPipeline_.headOfPipeline.flush();
   };
+
+  // Caption data has to be reset when seeking outside buffered range
+  this.resetCaptions = function() {
+    if (this.transmuxPipeline_.captionStream) {
+      this.transmuxPipeline_.captionStream.reset();
+    }
+  };
+
 };
 Transmuxer.prototype = new Stream();
 
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -157,14 +157,7 @@ var CaptionStream = function() {
     new Cea608Stream(1, 1) // eslint-disable-line no-use-before-define
   ];
 
-  this.activeCea608Channel_ = null;
-  // Since we don't know which channel is active until we get a control
-  // code that sets it, we start off with CEA608 handlers that just drop
-  // all the packets.
-  this.activeCea608Streams_ = [
-    packetDropper,
-    packetDropper
-  ];
+  this.reset();
 
   // forward data and done events from CCs to this CaptionStream
   this.ccStreams_.forEach(function(cc) {
@@ -199,8 +192,26 @@ CaptionStream.prototype.push = function(event) {
     return;
   }
 
+  // Sometimes, the same segment # will be downloaded twice. To stop the
+  // caption data from being processed twice, we track the latest dts we've
+  // received and ignore everything with a dts before that. However, since
+  // data for a specific dts can be split across 2 packets on either side of
+  // a segment boundary, we need to make sure we *don't* ignore the second
+  // dts packet we receive that has dts === this.latestDts_. And thus, the
+  // ignoreNextEqualDts_ flag was born.
+  if (event.dts < this.latestDts_) {
+    // We've started getting older data, so set the flag.
+    this.ignoreNextEqualDts_ = true;
+    return;
+  } else if ((event.dts === this.latestDts_) && (this.ignoreNextEqualDts_)) {
+    // We've received the last duplicate packet, time to start processing again
+    this.ignoreNextEqualDts_ = false;
+    return;
+  }
+
   // parse out CC data packets and save them for later
   this.captionPackets_ = this.captionPackets_.concat(parseCaptionPackets(event.pts, userData));
+  this.latestDts_ = event.dts;
 };
 
 CaptionStream.prototype.flush = function() {
@@ -240,6 +251,22 @@ CaptionStream.prototype.flush = function() {
     cc.flush();
   }, this);
   return;
+};
+
+CaptionStream.prototype.reset = function() {
+  this.latestDts_ = null;
+  this.ignoreNextEqualDts_ = false;
+  this.activeCea608Channel_ = null;
+  // Since we don't know which channel is active until we get a control
+  // code that sets it, we start off with CEA608 handlers that just drop
+  // all the packets.
+  this.activeCea608Streams_ = [
+    packetDropper,
+    packetDropper
+  ];
+  this.ccStreams_.forEach(function(ccStream) {
+    ccStream.reset();
+  });
 };
 
 CaptionStream.prototype.dispatchCea608Packet = function(packet) {
@@ -394,24 +421,7 @@ var Cea608Stream = function(field, dataChannel) {
   this.name_ = 'CC' + (((this.field_ << 1) | this.dataChannel_) + 1);
 
   this.setConstants();
-
-  this.mode_ = 'popOn';
-  // When in roll-up mode, the index of the last row that will
-  // actually display captions. If a caption is shifted to a row
-  // with a lower index than this, it is cleared from the display
-  // buffer
-  this.topRow_ = 0;
-  this.startPts_ = 0;
-  this.displayed_ = createDisplayBuffer();
-  this.nonDisplayed_ = createDisplayBuffer();
-  this.lastControlCode_ = null;
-
-  // Track row and column for proper line-breaking and spacing
-  this.column_ = 0;
-  this.row_ = BOTTOM_ROW;
-
-  // This variable holds currently-applied formatting
-  this.formatting_ = [];
+  this.reset();
 
   this.push = function(packet) {
     var data, swap, char0, char1, text;
@@ -615,6 +625,29 @@ Cea608Stream.prototype.flushDisplayed = function(pts) {
       stream: this.name_
     });
   }
+};
+
+/**
+ * Zero out the data, used for startup and on seek
+ */
+Cea608Stream.prototype.reset = function() {
+  this.mode_ = 'popOn';
+  // When in roll-up mode, the index of the last row that will
+  // actually display captions. If a caption is shifted to a row
+  // with a lower index than this, it is cleared from the display
+  // buffer
+  this.topRow_ = 0;
+  this.startPts_ = 0;
+  this.displayed_ = createDisplayBuffer();
+  this.nonDisplayed_ = createDisplayBuffer();
+  this.lastControlCode_ = null;
+
+  // Track row and column for proper line-breaking and spacing
+  this.column_ = 0;
+  this.row_ = BOTTOM_ROW;
+
+  // This variable holds currently-applied formatting
+  this.formatting_ = [];
 };
 
 /**
--- a/test/caption-stream.test.js
+++ b/test/caption-stream.test.js
@@ -13,7 +13,7 @@ var
 var makeSeiFromCaptionPacket = function(caption) {
   return {
     pts: caption.pts,
-    dts: caption.dts || caption.pts,
+    dts: caption.dts,
     nalUnitType: 'sei_rbsp',
     escapedRBSP: new Uint8Array([
       0x04, // payload_type === user_data_registered_itu_t_t35
@@ -35,6 +35,41 @@ var makeSeiFromCaptionPacket = function(caption) {
 
       0xff // marker_bits
     ])
+  };
+};
+
+// Create SEI nal-units from Caption packets
+var makeSeiFromMultipleCaptionPackets = function(captionHash) {
+  var pts = captionHash.pts,
+    dts = captionHash.dts,
+    captions = captionHash.captions;
+
+  var data = [];
+  captions.forEach(function(caption) {
+    data.push(0xfc | caption.type);
+    data.push((caption.ccData & 0xff00) >> 8);
+    data.push(caption.ccData & 0xff);
+  });
+
+  return {
+    pts: pts,
+    dts: dts,
+    nalUnitType: 'sei_rbsp',
+    escapedRBSP: new Uint8Array([
+      0x04, // payload_type === user_data_registered_itu_t_t35
+
+      (0x0b + (captions.length * 3)), // payload_size
+
+      181, // itu_t_t35_country_code
+      0x00, 0x31, // itu_t_t35_provider_code
+      0x47, 0x41, 0x39, 0x34, // user_identifier, "GA94"
+      0x03, // user_data_type_code, 0x03 is cc_data
+
+      // 110 00001
+      (0x6 << 5) | captions.length, // process_cc_data, cc_count
+      0xff // reserved
+    ].concat(data).concat([0xff /* marker bits */])
+    )
   };
 };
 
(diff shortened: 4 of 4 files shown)


squarebracket:
To explain the problem a little more, imagine you've got some caption data,

```
BOO!

                       AAAAAAAH!
```

That's been split up like this by the transcoder:

| Segment # | DTS | CC Data  |
|-----------|-----|----------|
|    ...    | ... |   ...    |
|     0     |  0  | `0x1370` |
|     0     |  1  |    BO    |
|     1     |  1  |    O!    |
|     1     |  2  | `0x147A` |
|     1     |  2  | `0x1723` |
|     1     |  3  |    AA    |
|     1     |  3  |    AA    |
|     2     |  3  |    AA    |
|     2     |  4  |    AH    |
|     2     |  4  |     !    |
|     2     |  5  |   EOC    |
|    ...    | ... |   ...    |

When VideoJS does an upswitch in quality, it'll grab the same segment # twice -- once for each rate. If the quality switch happens for segment 1, then the packets dispatched to the caption stream look like this:

| DTS | CC Data  |
|-----|----------|
| ... |   ...    |
|-----|----------|
|  0  | `0x1370` |
|-----|----------|
|  1  |    BO    |
|  1  |    O!    |
|-----|----------|
|  2  | `0x147A` |
|  2  | `0x1723` |
|-----|----------|
|  3  |    AA    |
|  3  |    AA    |
|-----|----------|
|  1  |    O!    |
|-----|----------|
|  2  | `0x147A` |
|  2  | `0x1723` |
|-----|----------|
|  3  |    AA    |
|  3  |    AA    |
|-----|----------|
|  3  |    AA    |
|-----|----------|
|  4  |    AH    |
|  4  |     !    |
|-----|----------|
|  5  |   EOC    |
|-----|----------|
| ... |   ...    |

where the dashed lines represent the division in packets as they're received by `CaptionStream`.

Since the whole caption is spread across two segments, we can't simply wait for the cue to finish; the caption data for segment 1 will be processed twice since the EOC is only in the second segment, and you'd get

```
BOO!

AAAAO!AAAAAAAH!
```

Conversely, you can't just reset the entire `CaptionStream` when you get the duplicate segment -- you'll chop off the beginning of the caption and end up with

```
O! AAAAAAAH!
```

So, you need some way of having the `CaptionStream` reject (or not be `push`ed) _just_ the packets from the duplicate segments, while still having the transmuxer do its thing with the video data.

My solution is to track the most recently received DTS. When a packet comes in that has a DTS less than that value, we know it's a repeated segment, so we drop the packet and set a flag. That flag is used when the incoming packet has a DTS equal to `latestDts`. In that last table above, you can see that DTS 3 is split into two packets -- one on either side of a segment boundary. We only want to ignore the second DTS == 3 packet (the repeated one), so if DTS == latestDts, and the flag is set, we clear the flag, drop the packet, and then resume normal processing.

The problem with this approach is that you need to wire up the transmuxer to reset the `CaptionStream`'s DTS tracking when there's a seek -- otherwise you get erroneously dropped packets when seeking backwards. I patched that into contrib-media-sources, and it worked, but I feel like maybe there's a better way to do all of this.

squarebracket:
@mjneil made a great suggestion about using `baseMediaDecodeTime`. No more extra triggering is required -- I simply call the `CaptionStream.resetPtsTracking()` from inside `setBaseMediaDecodeTime`. Works like a charm.

mjneil:
So it looks like relying on `baseMediaDecodeTime` will not be enough. `baseMediaDecodeTime` is called whenever we update the `timestampOffset` of the `SourceBuffer`.
```
// We will need to change timestampOffset of the sourceBuffer if either of
// the following conditions are true:
// - The segment.timeline !== this.currentTimeline
//   (we are crossing a discontinuity somehow)
// - The "timestampOffset" for the start of this segment is less than
//   the currently set timestampOffset
```
In a source with no discontinuities, this won't ever need to be called, so seeking backwards still causes caption data to be dropped.

What if instead of dropping all data that has `dts < lastSeenDts`, whenever we see `dts < lastSeenDts` keep all the parsed caption data you have that comes before `dts` and just throw out any parsed caption data that comes after `dts`, and just continue parsing the new data.
 
```
_ -> segment boundary 
seg 0a
BO
seg 1a
BO_O!AAAA
dupe seg1b -> toss out O!AAAA -> parse new seg1 data
BO_O!AAAA
seg2b
BO_O!AAAA_AAAH!
```

The only issue I see with this approach is that both `BO` and `O!` have the same dts value but split across segments. When you get to the condition of `dts < lastSeenDts` and need to throw out already parsed data, is there a way we can know to toss out `O!` but not `BO`

squarebracket:
Huh, I guess something was wrong with my initial test setup, because it did seem to work. But now that I try it again... it doesn't. Bummer.

> What if instead of dropping all data that has `dts < lastSeenDts`, whenever we see `dts < lastSeenDts` keep all the parsed caption data you have that comes before `dts` and just throw out any parsed caption data that comes after `dts`, and just continue parsing the new data.

There are problems with your solution as I understand it, but I'm not sure I understand it. By "keep all the parsed caption data", do you mean the state of the `Cea608Stream`? That, unfortunately, isn't possible. My example has teeny tiny segments, but in real life you can have a caption-ending control code in the middle of a segment, flushing the state of the `Cea608Stream` and pushing the cue up the transmuxer chain (attaching it to `this.pendingCaptions` I presume).

If your suggestion is to track which DTSs have already been processed by the `CaptionStream`, then you can make it work, but you also end up with a "how big should the array be?" question, and also you'd need to make sure to flush that information along with deleting the old video/audio buffer.

Ultimately, the goal here is to not process _specific_ packets twice, regardless of their DTS values. If it's possible / easy, probably the best way would be to simply toggle processing off when video.js sees that it's downloading the same time slice. i.e. it sees that a newly-downloaded segment's start and end overlap with the current buffer, and so it calls something like `CaptionStream.pauseProcessing()`, and then when it sees a newly-downloaded segment _doesn't_ overlap, it calls `CaptionStream.resumeProcessing`. Essentially, the `CaptionStream` having knowledge of whether it's a duplicate segment, or something else that does know telling the `CaptionStream`, is required for an elegant solution.

Without that information, I think all you can really do is ignore packets with `dts < lastSeenDts`. The `CaptionStream` only knows the information of the packets it processes, and since you can have any number of packets in a segment, has no information regarding the segment boundaries. If it can know at least when to _start_ ignoring packets, then you're in business, because you have a window for which you should drop packets. Otherwise I think you have to just go with my solution, where things are reset on seek.

Anyway, I marked this PR as WIP since I only really know the specific caption parts, and not the general video.js architecture. It's possible there's a better solution elsewhere in the HLS pipeline.

mjneil:
I see what you mean about the caption-ending control in the middle of the segment sending the captions off to the `CoalesceStream.pendingCaptions`.

I still think this basic principle can work with some variation. 

So it looks like in a sense, `CaptionStream`s state is "reset" for every segment. We always `push` a segment, and then `flush` it. So `CaptionStream.captionPackets_` is always cleared after each segment, but the `Cea608Stream.displayed_` may hold onto some of the caption data depending on if it receives control codes or not to end the caption by calling `Cea608Stream.flushDisplayed`.  When `flushDispalyed` is called, everything being held onto gets combined to a caption and sent off to `CoalesceStream.pendingCaptions`.

Here is what I propose. Continue to track `CaptionStream.latestDts_`, however, when you come across an `event.dts < this.latestDts_` instead of ignoring the packet, flush what is currently in the `Cea608Stream`s effectively sending off an incomplete caption to `CoalesceStream.pendingCaptions` and reset those stream's state so their `Cea608Stream.startPts_` is the repeated `event.dts` you received. Also reset`CaptionStream.latestDts_` to this repeated `event.dts` and continue processing it as normal. Basically `CaptionStream` and `Cea608Stream` will process everything it gets. 

Now to prevent the duplication. I think that `CoalesceStream` could start tracking a `latestCaptionStartPts_` as well (it may be sufficient to just use the pts of the last caption in `CoalesceStream.pendingCaptions`).  Whenever the `CoalesceStream` receives a new caption, it should compare that `caption.startPts` to `latestCaptionStartPts`. If `caption.startPts < latestCaptionStartPts` then ignore the caption and do nothing.  If `caption.startPts === latestCaptionStartPts` then remove the caption from `pendingCaptions` that has `latestCaptionStartPts` and push the new one you received as a replacement. `latestCaptionStartPts` will be reset on `flush`.

I think as a last measure to prevent duplicates after rendition switches, when contrib-media-sources  receives captions from the transmuxer, it will have to first remove captions from the text track that overlap with the captions it just received.

Does this make sense and sound possible?

mjneil:
One concern I have with my proposal is the flushing of an incomplete caption in `Cea608Stream` since it looks like there are different codes that cause a flushing of the caption, and each updates the state of the stream in a different way. So I'm not sure which would be the proper way to reset the state in this case, or if it is possible to do

squarebracket:
Changing the state of a `Cea608Stream` is pretty dangerous, IMO. For one thing, your captions can be like this:

seg 1 --> control code for start data
seg 2 --> just text data for the caption
seg 3 --> control code that causes flush

Now, imagine that segment 2 is your repeated segment. Resetting the `Cea608Stream` means you now have a botched caption, because you've lost the data from segment 1.

I really think the right approach here is simply dropping duplicate data. The spec was made for constant transmission, and hacking in extra logic to that could lead to undesirable behaviour.

The `CaptionStream` doesn't receive segments, it only receives packets (assuming you're referring to `CaptionStream.push`). Perhaps, earlier on in the transmuxer chain, we can track duplicate segments, before the packets are unwrapped and separately dispatched. If we can get the starting DTS and ending DTS of a segment, then we can simply drop all duplicate caption DTS within the time range (except, you have to let through the second `lastDts` you see, like I'm doing in this PR).

I tried looking into this, but it seemed a bit too complicated. Maybe I should revisit that possibility.

mjneil:
Hmmm that is problematic.

If knowing the start and end dts of the segment is enough information, I think we can get that easily. 

You're right that the `CaptionStream` only receives packets and not segments. However, we can use `push` and `flush` to help us detect segment boundaries.

The usage of the transmuxer is 
```
transmuxer.push(segment);
transmuxer.flush();
```
Since everything is happening synchronously, the data will make it through the full `push` pipeline 
before calls to `flush` and respectively another call to push won't start until the `flush` pipeline has completed.  You could use a boolean to determine if the call to `push` or `flush` to mark your segment boundaries.

I think that seeking backwards will still be an issue. Without a way to notify the transmuxer a seek happened, it will think that the new segment it is getting is a duplicate and drop it. While it seems less than ideal, I don't think it would be too bad for the `HtmlMediaSource` to listen for seek events and send a message to the transmuxer to reset caption dts

squarebracket:
Hmm. I had an idea about the transmuxer tracking the starting and ending DTS of the most recently parsed segment `n`, but I think that doesn't work in all cases. Like if you've buffered up more than a segment's worth of data, then switch quality, the new segment could be `n-1` or `n-2`.

So yeah, I guess the problem is actually "it should drop captions if that DTS is in the buffer" rather than the "it should drop captions if it receives the same segment twice." In which case, I guess the seek event really is the easiest way of doing this :\

mjneil:
You're right, its hard to know which segment will be next depending on how much we've buffered forward. 

I think it will be ok to expose an api on the transmuxer for resetting caption stream pts tracking on a seek and contrib-media-sources calls that api on seek events. Would you mind creating another PR for contrib-media-sources with the seek patch?

squarebracket:
Sure, no problem.

lib/mp4/transmuxer.js
```diff
@@ -1248,6 +1249,11 @@ Transmuxer = function(options) {
     // Start at the top of the pipeline and flush all pending work
     this.transmuxPipeline_.headOfPipeline.flush();
   };
+
+  this.seeked = function() {
+    this.transmuxPipeline_.captionStream.resetDtsTracking();
```
mjneil: This should probably be wrapped in an `if (this.transmuxPipeline_.captionStream)` since `captionStream` won't exist if an AAC pipeline is setup.

lib/mp4/transmuxer.js
```diff
@@ -1213,6 +1213,7 @@ Transmuxer = function(options) {
       videoTrack.timelineStartInfo.dts = undefined;
       videoTrack.timelineStartInfo.pts = undefined;
       clearDtsInfo(videoTrack);
+      pipeline.captionStream.resetDtsTracking();
```
squarebracket: Should I keep this, if we're triggering the reset on a `seek`?

mjneil: Yea this also gets called when contrib-hls detects a discontinuity, which does not require a seek to happen. You'll still want to reset the caption dts tracking on discontinuities.

squarebracket: Ok thanks.

lib/mp4/transmuxer.js
squarebracket: done.

mjneil:
Would you be able to add a simple unit test that follows the overlap segment example and outputs a proper caption thanks to the dts tracking as well as showing it will being processing captions again after simulating a seek backwards?

squarebracket:
Of course, I haven't added any unit tests yet. Was waiting until we resolved the proper approach. Now that that's done with, I'll code them up.

squarebracket:
Ok, so when I started writing up the tests, I realized that probably some other stuff should be reset when seeking, like the buffers in the `Cea608Stream`s. Now, I don't remember ever getting a caption mashed together with another after a seek. Maybe I was lucky, or maybe there's something else somewhere in the code that has the side effect of generally preventing this. But, I figured that since we're hooking into a seek anyway, might as well just recode things so that everything is reset on a seek. That way we're starting fresh w/r/t captions.

I've made a few unit tests to test the proper behaviour. Let me know if you have any suggestions for other tests. I didn't add a test to see if `reset()` was fired when calling `transmuxer.seeked()`, but I feel like that's unnecessary? I think probably the best place for that would be a test in contrib-media-sources to make sure that `seeked` is fired.

mjneil:
I think you might experience some problems when doing an in-buffer seek. If you buffer up 30 seconds of content, and seek 15 seconds forward, nothing really gets reset in hls/mux pipeline. So i'm imagining the last segment you appended had the start of a caption but not complete. (the next segment will complete it). Now you do a seek 15 seconds forward into the buffer. nothing needs to be reset because HLS will just be grabbing and appending the next segment. Before when only reseting the dts tracking, this would have no problems because it would just start tracking the new dts value, but the incomplete caption it parsed and held onto is still there. With the update to reset everything, state will be reset, and then we append the next segment, and you lose the first half of the caption

squarebracket:
Hmm, yes good point. Interestingly, I did just get a botched (incomplete) caption when doing some randon seeks on a version without the reset.

mjneil:
Hmmm could it be a combination of the seek and overlapping duplicate segments. Basically same idea in my previous comment, but instead of the next segment being appended  its a duplicate because of a rendition switch happening at the same time and that segment causes the one "in progress" to be flushed and sent out incomplete

squarebracket:
Hmm, tough to say. I think I'd have to print the actual data that's received by the `Cea608Stream` to have a good idea of what's going on.

Alternatively, I could simply have contrib-media-sources pass something like `inBuffer` to let the transmuxer know what should be reset.

mjneil:
You may not have to pass information in and only call the seek reset api from contrib-media-sources when it was an out of buffer seek

squarebracket:
Ah yeah, true -- since the captions will already be processed for an in-buffer seek.

lib/mp4/transmuxer.js
```diff
@@ -1213,6 +1213,7 @@ Transmuxer = function(options) {
       videoTrack.timelineStartInfo.dts = undefined;
       videoTrack.timelineStartInfo.pts = undefined;
       clearDtsInfo(videoTrack);
+      pipeline.captionStream.reset();
```
mjneil: I would wrap this in an `if (pipeline.captionStream)`

squarebracket: I don't believe that's necessary, since if there's a `videoTrack` then there must be a `captionStream`

mjneil: ah ya should be good then 👍

squarebracket:
Got one more test to add before this is merged in.

squarebracket:
Alright, squashed and ready to go :+1:

lib/flv/transmuxer.js
```diff
@@ -400,6 +400,13 @@ Transmuxer = function(options) {
     packetStream.flush();
   };
 
+  // Caption data has to be reset when seeking outside buffered range
+  this.resetCaptions = function() {
+    if (this.transmuxPipeline_.captionStream) {
```
mjneil: looks like the flv transmuxer doesnt have a `this.transmuxPipeline_` this can just be updated to
```
this.resetCaptions = function() {
  captionStream.reset();
};
```

squarebracket: Oops, yup, you're totally right.

Issue #106, opened 2016-08-08
Problems with multiple track audio.


londono:
Hi, first good job with this mux.js. Second, I have a chunk of 10 sec of .ts with 1 Video track and 3 Audio tracks and I am using your debug site. I have one error : mux.js line 4090 Uncaught TypeError: Cannot read property '0' of undefined.  The same kind of video with 1 video 1 Audio is working just fine. Can you help me?, please!

BrandonOCasey:
Can you post a sample of the TS that is giving you an issue?

londono:
1 Video track and 3 Audio tracks 
[segments1.zip](https://github.com/videojs/mux.js/files/407044/segments1.zip)

londono:
My own mistake in the creation of the segment. Now I can read the segment TS but the audio output seems to be a mix of two streams audio.  I got a video TS 1 video and 2 audios (when the audio is just the same stream but repeated).  How I can  manipule the tracks??. here is my segment 

[testTwoAudios.zip](https://github.com/videojs/mux.js/files/413520/testTwoAudios.zip)

synologic:
Same issue here (actually coming from hls.js). If multiple audio tracks are present, all are sent to the transmuxer, resulting in garbled audio or even video crash (in chrome). Would be helpful to have an audio selection based on the PMT content, or at least filter all but one audio track (which i did in my case).

cronello:
Same issue with multiple audio tracks when using videojs-contrib-hls.
At `ElementaryStream` is just one buffer for video, audio, so when multiple audio track came all audio is pushed into the same audio buffer.
Just for my workaround, I push into audio buffer just the first audio track, by PID from PMT table
at `ElementaryStream.push pes function` I replace `case StreamTypes.ADTS_STREAM_TYPE` content by:
```javascript
case StreamTypes.ADTS_STREAM_TYPE:
            // take care about the first audio track and not mixing multiple audio 
            // into one output stream
            if (!audio.pid) {
                audio.pid = data.pid;
            }
            if (audio.pid === data.pid) {
                stream = audio;
                streamType = 'audio';
                break;
            }
            // ignor other audio track
            return;
```

I'll be good, to have some API for selecting audio track

imbcmdth:
@cronello This change would be great to have a PR for. Thanks for looking into this.

We don't have any plans to support multiple audio tracks in a segment but we definitely should still pull out a single track (even if multiple are present) correctly.

mjneil:
[...]

PR #164, opened 2017-10-06
Use the first audio and video tracks in the PMT

mjneil:
Segments can contain multiple audio and video tracks. Currently mux.js does not take this into account, so when there are multiple, they get combined into a single stream. This can cause playback issues or distorted audio. 
https://github.com/videojs/videojs-contrib-hls/issues/994
https://github.com/videojs/mux.js/issues/106
https://github.com/videojs/videojs-contrib-hls/issues/1247

This change makes sure the transmuxer uses the track with the lowest identifier, ignoring extra tracks.

I'm not quite sure if it is common for streams to have multiple timed-metadata tracks, and we have not gotten any reports around multiple timed-metadata tracks, so I have kept the functionality of combining all timed-metadata tracks into a single stream.

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -156,7 +156,11 @@ TransportParseStream = function() {
     }
 
     // overwrite any existing program map table
-    self.programMapTable = {};
+    self.programMapTable = {
+      video: null,
+      audio: null,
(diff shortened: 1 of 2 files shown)


forbesjo:
Looks good to me

PR #167, opened 2017-10-23
Set active data channel per-field instead of globally

squarebracket:
According to some real-world caption data for CC4, the currently active data channel is tracked on a per-field basis, not globally for all captions.

Comments about the efficiency of my solution would be particularly appreciated.

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -260,9 +260,9 @@ CaptionStream.prototype.reset = function() {
 
 CaptionStream.prototype.dispatchCea608Packet = function(packet) {
   // NOTE: packet.type is the CEA608 field
-  if (this.setsChannel1Active(packet) && this.activeCea608Channel_[packet.type] !== 0) {
+  if (this.setsChannel1Active(packet)) {
     this.activeCea608Channel_[packet.type] = 0;
-  } else if (this.setsChannel2Active(packet) && this.activeCea608Channel_[packet.type] !== 1) {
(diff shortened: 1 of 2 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -256,30 +252,25 @@ CaptionStream.prototype.flush = function() {
...
 
 CaptionStream.prototype.dispatchCea608Packet = function(packet) {
-  if (this.setsChannel1Active(packet) && this.activeCea608Channel_ !== 1) {
-    this.activeCea608Channel_ = 1;
-    this.activeCea608Streams_ = [this.ccStreams_[0], this.ccStreams_[2]]; // CC1, CC3
-  } else if (this.setsChannel2Active(packet) && this.activeCea608Channel_ !== 2) {
-    this.activeCea608Channel_ = 2;
-    this.activeCea608Streams_ = [this.ccStreams_[1], this.ccStreams_[3]]; // CC2, CC4
+  // NOTE: packet.type is the CEA608 field
+  if (this.setsChannel1Active(packet) && this.activeCea608Channel_[packet.type] !== 0) {
```
mjneil: is the addition of ` && this.activeCea608Channel_[packet.type] !== 0` still necessary here? It doesn't seem like there would be any harm in setting `this.activeCea608Channel_[packet.type] = 0;` if it already is 0 (same for the check below)

squarebracket: Yup, I suppose that's true.

PR #174, opened 2018-01-05
Add option to keep original timestamps in MP4 transmuxer.

### LINKED ISSUE #168: Can mux.js be configured not to change timestamps?
We use mux.js, if present, to convert TS to MP4 in Shaka Player on platforms that don't have native TS support.  However, when we do that, we've already parsed the timestamps from TS segments to build an accurate segment index.  What we find, though, is that the timestamps in the output are not the same as the input timestamps.  See google/shaka-player#1102 for more.

To support both native TS implementations and mux.js-based transmuxing to MP4, we need everything to be consistent.  For our architecture, the timestamps in the MP4 output should be the same as the TS input segments.

Can this be done today with mux.js?  Or would this constitute an enhancement?

TheModMaker:
Closes #168

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -120,6 +120,7 @@ sumFrameByteLengths = function(array) {
  * Constructs a single-track, ISO BMFF media segment from AAC data
  * events. The output of this stream can be fed to a SourceBuffer
  * configured with a suitable initialization segment.
+ * @param track {object} track metadata configuration
  * @param options {object} transmuxer options object
  * @param options.keepOriginalTimestamps {boolean} If true, keep the timestamps
  *        in the source; false to adjust the first segment to start at 0.
@@ -527,6 +528,10 @@ VideoSegmentStream = function(track, options) {
 
     // Continue with the flush process now
     this.trigger('done', 'VideoSegmentStream');
+  };
+
+  this.setOptionsForTesting_ = function(newOptions) {
+    options = newOptions;
   };
 
   this.resetStream_ = function() {
@@ -986,6 +991,9 @@ clearDtsInfo = function(track) {
  * Calculate the track's baseMediaDecodeTime based on the earliest
  * DTS the transmuxer has ever seen and the minimum DTS for the
  * current track
+ * @param track {object} track metadata configuration
+ * @param keepOriginalTimestamps {boolean} If true, keep the timestamps
+ *        in the source; false to adjust the first segment to start at 0.
  */
 calculateTrackBaseMediaDecodeTime = function(track, keepOriginalTimestamps) {
   var
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -120,14 +120,19 @@ sumFrameByteLengths = function(array) {
  * Constructs a single-track, ISO BMFF media segment from AAC data
  * events. The output of this stream can be fed to a SourceBuffer
  * configured with a suitable initialization segment.
+ * @param options {object} transmuxer options object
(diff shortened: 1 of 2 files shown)


TheModMaker:
FYI @joeyparrish

lib/mp4/transmuxer.js
```diff
@@ -120,15 +120,20 @@ sumFrameByteLengths = function(array) {
  * Constructs a single-track, ISO BMFF media segment from AAC data
  * events. The output of this stream can be fed to a SourceBuffer
  * configured with a suitable initialization segment.
+ * @param options {object} transmuxer options object
```
joeyparrish: Looks like this was already missing the param annotation for "track".  You could maybe copy this one from VideoSegmentStream:

```js
  * @param track {object} track metadata configuration
```

lib/mp4/transmuxer.js
```diff
@@ -978,20 +987,23 @@ clearDtsInfo = function(track) {
  * DTS the transmuxer has ever seen and the minimum DTS for the
  * current track
  */
-calculateTrackBaseMediaDecodeTime = function(track) {
+calculateTrackBaseMediaDecodeTime = function(track, keepOriginalTimestamps) {
```
joeyparrish: Document params here, too?  I know track docs were already missing, but it wouldn't hurt to add them here for both params.

joeyparrish:
@mjneil, any feedback you have would be most welcome.  Thanks!

mjneil:
This is looking good. Would you be able to add a simple test that uses this option? Can probably do something very similar to one the of `baseMediaDecodeTime` tests here https://github.com/videojs/mux.js/blob/master/test/transmuxer.test.js#L2040

TheModMaker:
Added a test and added docs like Joey suggested.

test/transmuxer.test.js
```diff
@@ -2074,6 +2074,45 @@ QUnit.test('calculates baseMediaDecodeTime values from the first DTS ever seen a
   QUnit.equal(tfdt.baseMediaDecodeTime, 90, 'calculated baseMediaDecodeTime');
 });
 
+QUnit.test('doesn\'t adjust baseMediaDecodeType when configured to', function() {
+  videoSegmentStream.setOptionsForTesting_({keepOriginalTimestamps: true});
```
mjneil: I think it would be better if we instead updated the QUnit `setup` function for this module (line 1425) to add an options object and then alter that object here instead of adding an api for testing to the actual source.
e.g.
```
var track = {};
var options = {};
videoSegmentStream = new VideoSegmentStream(track, options);
videoSegmentStream.options = options;
videoSegmentStream.track = track;
.....
```
and then here just 
```
videoSegmentStream.options.keepOriginalTimestamps = true;
```

TheModMaker: Done

test/transmuxer.test.js
```diff
@@ -2074,6 +2076,45 @@ QUnit.test('calculates baseMediaDecodeTime values from the first DTS ever seen a
   QUnit.equal(tfdt.baseMediaDecodeTime, 90, 'calculated baseMediaDecodeTime');
 });
 
+QUnit.test('doesn\'t adjust baseMediaDecodeType when configured to', function() {
```
gesinger: Minor: s/baseMediaDecodeType/baseMediaDecodeTime/, and maybe specify "when configured to keep original timestamps"

TheModMaker: Done

forbesjo:
The build is fixed now and I reran the tests

PR #175, opened 2018-01-14
Simplify .travis config

forbesjo:
Headless Chrome should take care of everything. Also trying to fix the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,5 @@ node_js:
   - 'node'
   - 'lts/argon'
 addons:
-  # TODO: This should be using latest for firefox. It is currently turned off because
-  #       of https://github.com/travis-ci/travis-ci/issues/8242 When this issue is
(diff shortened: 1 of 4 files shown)


squarebracket:
Is there a reason why there's no firefox config? According to [this](https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-the-Firefox-addon-in-headless-mode) travis should support a headless Firefox, and [the karma launcher](https://github.com/karma-runner/karma-firefox-launcher/blob/master/karma.conf.js) knows of a `FirefoxHeadless` browser.

forbesjo:
I'll add Firefox to the test config, previously we were having issues with it in Travis.

squarebracket:
Ah yeah true, I do remember that being a thing. Seems to be good now, hopefully it doesn't break again :crossed_fingers:

PR #181, opened 2018-03-01
Get frame duration when inspecting segments with audio

gesinger:


lib/m2ts/probe.js
```diff
@@ -113,6 +114,28 @@ var parsePesType = function(packet, programMapTable) {
...
+    i,
+    offset = 4 + parseAdaptionField(packet);
+
+  if (offset >= packet.byteLength) {
+    return null;
+  }
+
+  // check for the ADTS header syncword 0xFFF
+  for (i = offset; i < packet.length - 1; i++) {
+    if (packet[i] === 0xFF && packet[i + 1] >> 4 === 0x0F) {
```
mjneil: any reason in switching to the bitshift here instead of the inverse of the check `AdtsStream` does? `(buffer[i + 1] & 0xF6) !== 0xF0`

lib/tools/ts-inspector.js
```diff
@@ -79,6 +83,95 @@ var parsePsi_ = function(bytes, pmt) {
...
+    i,
+    crcPresent,
+    headerLength,
+    nextPacketResult,
+    packetsAvailable = (bytes.byteLength - startIndex - 1) / MP2T_PACKET_LENGTH;
+
+  for (i = 0; i < packetsAvailable; i++) {
+    headerOffset = probe.ts.getAdtsHeaderOffset(packet);
+
+    if (headerOffset) {
```
mjneil: I think this has the potential to break for headers split across pes packets. From what I can tell, if the header is split across packets, the syncword will only be in the first packet (or split itself), so the second half of the packet may or may not get appended depending on if any of the values in the packet satisfy the sync word check.

lib/m2ts/probe.js
gesinger: I just felt it was easier to read. Do you have a preference?

mjneil: I am leaning towards reverting back to the original check. From https://wiki.multimedia.cx/index.php?title=ADTS the first two bytes of the ADTS header have the format `AAAAAAAA AAAABCCD`

>B | 1 | MPEG Version: 0 for MPEG-4, 1 for MPEG-2
>C | 2 | Layer: always 0
>D | 1 | protection absent, Warning, set to 1 if there is no CRC and 0 if there is CRC

The `0xF6` is making sure that those two `C` bits are `0`, where as your check ignores those bits. I think it will be safer to make sure these bits are checked as well.

lib/tools/ts-inspector.js
```diff
@@ -121,21 +121,32 @@ var nextPesPacket_ = function(bytes, pmt, startIndex, type) {
...
+    // elementary stream bytes
+    esBytes,
+    packetEsBytes,
+    pesHeaderLength,
+    headerBytes,
+    // to get the elementary stream bytes from the packet, we add 4 for the TS header,
+    // then since we know it is an audio PES packet, we can bypass the 4 byte start code,
+    // 2 byte PES packet length, and the first 2 optional PES header bytes to get to the
+    // PES header length (remainder) byte
+    pesHeaderLengthIndex = 4 + 4 + 2 + 2,
```
mjneil: This doesn't take into account the adaption field of the TS packet header. https://github.com/videojs/mux.js/blob/master/lib/m2ts/m2ts.js#L222

lib/tools/ts-inspector.js
```diff
@@ -121,21 +121,32 @@ var nextPesPacket_ = function(bytes, pmt, startIndex, type) {
...
     packetsAvailable = (bytes.byteLength - startIndex - 1) / MP2T_PACKET_LENGTH;
 
   for (i = 0; i < packetsAvailable; i++) {
-    headerOffset = probe.ts.getAdtsHeaderOffset(packet);
+    pesHeaderLength = packet[pesHeaderLengthIndex];
+    packetEsBytes = packet.subarray(pesHeaderLengthIndex + pesHeaderLength + 1, endIndex);
```
mjneil: a comment mentioning what the `+ 1` is for here would be helpful

forbesjo:
Looks good to me

forbesjo:
Let's review this after the transmux before append work and see if we still need this.

gkatsev:
Closing for now. TBA work makes this unnecessary for VHS, for others, it may not make a huge difference as it's only one frame duration shorter. This does make segment metadata in VHS a bit more complicated especially since Video.js's activeCues logic removes things on >= rather than on >.

gkatsev:
This may be something we want to fix in the future but we'll want to coordinate with the fix in Video.js and everywhere else.

PR #185, opened 2018-03-14
fix mp4 probe bitwise operations

mjneil:
javascript bitwise operations convert operands to signed 32 bit integers, this, the output of bitwise ops will be signed 32bit ints, however, we want them to be unsigned. `>>> 0` ensures the value will be unsigned

### DIFF
--- a/test/utils.bin.test.js
+++ b/test/utils.bin.test.js
@@ -1,0 +1,26 @@
+var
+  QUnit = require('qunit'),
+  toUnsigned = require('../lib/utils/bin').toUnsigned;
+
+QUnit.module('Binary Utils');
+
+QUnit.test('converts values to unsigned integers after bitwise operations', function() {
+var bytes;
+
+bytes = [0, 0, 124, 129];
+
+QUnit.equal(toUnsigned(bytes[0] << 24 |
+                       bytes[1] << 16 |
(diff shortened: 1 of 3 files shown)


forbesjo:
Confirmed that this fixes up https://github.com/videojs/http-streaming/pull/48

test/utils.bin.test.js
```diff
@@ -0,0 +1,26 @@
+var
+  QUnit = require('qunit'),
+  toUnsigned = require('../lib/utils/bin').toUnsigned;
+
+QUnit.module('Binary Utils');
+
+QUnit.test('converts values to unsigned integers after bitwise operations', function() {
+var bytes;
```
gesinger: Minor: indentation

test/utils.bin.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+QUnit.equal(toUnsigned(bytes[0] << 24 |
+                       bytes[1] << 16 |
+                       bytes[2] <<  8 |
+                       bytes[3]),
+            31873, 'positive signed result stays positive');
+
+bytes = [150, 234, 221, 192];
+
+// The following bitwise operation would normally result in -1762992704
```
gesinger: Can also prove it by doing it without the toUnsigned

Issue #187, opened 2018-03-28
remuxed video play at IE11 or Edge, show afterimage


purpletortoise-choi:
After i do remux ts to mp4 at Edge or IE 11 (windows10).
I try to play remuxed video. 

Remuxed video play well, but show afterimage when backward or forward seeking.

![afterimageissue](https://user-images.githubusercontent.com/11069488/38021242-4e747544-32b7-11e8-8131-4fb78c1eab3e.gif)

Sample Url is 'https://video-dev.github.io/streams/x36xhzz/url_6/url_846/193039199_mp4_h264_aac_hq_7.ts'

Please check it.
Thanks

mjneil:
Edge and IE11 seem to have some problems with their MSE implementation. I've reported a similar issue to Microsoft here https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14600417/

elasim:
safari browser has same issue

PR #188, opened 2018-03-30
fix: remuxed video corruption fix at ie, edge

purpletortoise-choi:
I registered the issue #187 .

Fixed code in transmuxer.js file.
I added ```isNonSyncSample``` property at the createDefaultSample function used by ```VideoSegmentStream```.

```isNonSyncSample``` property semantics by MP4_spec;
"The flag sample_is_non_sync_sample provides the same information as the sync sample table [8.6.2].
When this value is set 0 for a sample, it is the same as if the sample were not in a movie fragment and
marked with an entry in the sync sample table (or, if all samples are sync samples, the sync sample table
were absent)."

```isNonSyncSample``` property is already used by the mp4-generator.js file.
```js
  videoTrun = function(track, offset) {
    var bytes, samples, sample, i;

    samples = track.samples || [];
    offset += 8 + 12 + (16 * samples.length);

    bytes = trunHeader(samples, offset);

    for (i = 0; i < samples.length; i++) {
      sample = samples[i];
      bytes = bytes.concat([
         ...
        (sample.flags.isLeading << 2) | sample.flags.dependsOn,
        (sample.flags.isDependedOn << 6) |
          (sample.flags.hasRedundancy << 4) |
          (sample.flags.paddingValue << 1) |
          sample.flags.isNonSyncSample,
        sample.flags.degradationPriority & 0xF0 << 8,
        sample.flags.degradationPriority & 0x0F, // sample_flags
         ...
      ]);
    }
    return box(types.trun, new Uint8Array(bytes));
  };
```

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -63,7 +63,8 @@ createDefaultSample = function() {
       dependsOn: 1,
       isDependedOn: 0,
       hasRedundancy: 0,
-      degradationPriority: 0
+      degradationPriority: 0,
+      isNonSyncSample: 1
     }
   };
 };
@@ -756,6 +757,7 @@ VideoSegmentStream = function(track, options) {
 
         if (currentFrame.keyFrame) {
           sample.flags.dependsOn = 2;
+          sample.flags.isNonSyncSample = 0;
         }
 
         dataOffset += sample.size;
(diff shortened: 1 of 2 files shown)


mjneil:
@purpletortoise-choi Thanks for the PR! This seems to work really well. Would you be able to extend one of the tests to have an assertion that the flag gets added to key frame samples?

purpletortoise-choi:
@mjneil Thanks for your comment. I add test code about ```isNonSyncSample``` flag validate.
Please check test code. 
Thank you. : )

forbesjo:
Looks good to me, merging

PR #196, opened 2018-06-11
add some docs and a couple small test changes

squarebracket:
When going through caption stuff with @ldayananda, I noticed that several of the tests I committed were missing helpful comments for the control codes. In that process, I noticed a couple tests that were doing some dumb things, so now they are no longer doing dumb things.

Also realized that there was no test for special/extended characters for CC2-CC4, so I added a small test. Full character mapping functionality is tested for `Cea608Stream`, but only for CC1.

### DIFF
--- a/test/caption-stream.test.js
+++ b/test/caption-stream.test.js
@@ -872,7 +872,8 @@ QUnit.test('ignores CEA708 captions', function() {
 });
 
 // Full character translation tests are below for Cea608Stream, they just only
-// test support for CC1
+// test support for CC1. See those tests and the source code for more about the
+// mechanics of special and extended characters.
 QUnit.test('special and extended character codes work regardless of field and data channel', function() {
   var packets, seiNals, captions = [];
   packets = [
--- a/test/caption-stream.test.js
+++ b/test/caption-stream.test.js
@@ -871,21 +871,32 @@ QUnit.test('ignores CEA708 captions', function() {
   QUnit.equal(captions[2].text, 'WE TRY NOT TO PUT AN ANIMAL DOWN\nIF WE DON\'T HAVE TO.', 'parsed third caption correctly');
 });
 
-// Full character translation testa are below for Cea608Stream, they just only
+// Full character translation tests are below for Cea608Stream, they just only
 // test support for CC1
 QUnit.test('special and extended character codes work regardless of field and data channel', function() {
   var packets, seiNals, captions = [];
   packets = [
+    // RU2 (roll-up, 2 rows), CC2
     { ccData: 0x1c25, type: 0 },
+    // ®
     { ccData: 0x1930, type: 0 },
+    // CR (carriage return), CC2, flush caption
     { ccData: 0x1c2d, type: 0 },
+    // RU2, CC3
     { ccData: 0x1525, type: 1 },
+    // "
     { ccData: 0x2200, type: 1 },
+    // «
     { ccData: 0x123e, type: 1 },
+    // CR, CC3, flush caption
     { ccData: 0x152d, type: 1 },
+    // RU2, CC4
     { ccData: 0x1d25, type: 1 },
+    // "
     { ccData: 0x2200, type: 1 },
+    // »
     { ccData: 0x1a3f, type: 1 },
+    // CR, CC4, flush caption
     { ccData: 0x1d2d, type: 1 }
   ];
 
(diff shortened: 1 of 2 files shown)


test/caption-stream.test.js
```diff
@@ -671,38 +675,54 @@ QUnit.test('drops data until first command that sets activeChannel for a field',
...
     { pts: 4 * 1000, ccData: characters('el'), type: 0 },
     { pts: 4 * 1000, ccData: characters('d0'), type: 0 },
+    // EOC, mark display start
+    { pts: 5 * 1000, ccData: 0x142f, type: 0 },
+    // EOC, duplicated as per spec
     { pts: 5 * 1000, ccData: 0x142f, type: 0 },
-    { pts: 5 * 1000, ccData: 0x1420, type: 0 },
+    // EOC, mark display end and flush
     { pts: 6 * 1000, ccData: 0x142f, type: 0 },
+    // EOC not duplicated cuz not necessary
```
ldayananda: `cuz`? 🙃 Is the reason it's not necessary because it's the 3rd EOC?

test/caption-stream.test.js
```diff
@@ -815,6 +871,37 @@ QUnit.test('ignores CEA708 captions', function() {
   QUnit.equal(captions[2].text, 'WE TRY NOT TO PUT AN ANIMAL DOWN\nIF WE DON\'T HAVE TO.', 'parsed third caption correctly');
 });
 
+// Full character translation testa are below for Cea608Stream, they just only
```
ldayananda: typo `testa`

test/caption-stream.test.js
```diff
@@ -815,6 +871,37 @@ QUnit.test('ignores CEA708 captions', function() {
...
+
+  captionStream.on('data', function(caption) {
+    captions.push(caption);
+  });
+
+  seiNals = packets.map(makeSeiFromCaptionPacket);
+  seiNals.forEach(captionStream.push, captionStream);
+  captionStream.flush();
+
+  QUnit.deepEqual(captions[0].text, String.fromCharCode(0xae), 'CC2 special character correct');
```
ldayananda: Maybe add a comment with what the special/extended character is?

test/caption-stream.test.js
```diff
@@ -624,28 +625,31 @@ QUnit.test('extracts all theoretical caption channels', function() {
...
-    { pts: 13000, type: 0, ccData: characters('1b') },
-    { pts: 14000, type: 0, ccData: characters('1c') },
-    { pts: 15000, type: 0, ccData: 0x142d },
-    { pts: 16000, type: 1, ccData: 0x152d },
-    { pts: 17000, type: 1, ccData: 0x1d2d },
-    { pts: 18000, type: 0, ccData: 0x1c25 },
-    { pts: 19000, type: 0, ccData: characters('2b') },
-    { pts: 20000, type: 0, ccData: 0x1c2d }
+    // instead of globally
+    { pts: 6000, type: 0, ccData: characters('2a') }, // CC2
```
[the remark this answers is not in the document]
squarebracket: No. This is field 0, and the last control code (`0x1c25`) set `CC2` active. `CC3` is field 1.

test/caption-stream.test.js
squarebracket: It's not necessary because duplicate control codes are simply ignored (as opposed to waiting for the duplicated control code to process it). We only need to duplicate a control code if it's used twice in a row (e.g. to flush a pop-on caption). I could duplicate all control codes always in all tests, but I feel like that would be overkill.

test/caption-stream.test.js
squarebracket: :+1: thanks for catching that

test/caption-stream.test.js
squarebracket: Added more comments.

test/caption-stream.test.js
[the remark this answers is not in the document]
ldayananda: Oh, missed that. Thanks!

test/caption-stream.test.js
ldayananda: Oh yeah, I see. I agree the comment is probably enough

test/caption-stream.test.js
ldayananda: One last thing that would be good to mention is that we expect `"` to be deleted because they come before an extended character

ldayananda:
One last thing that would be nice to spell out, but other than that LGTM.

test/caption-stream.test.js
squarebracket: I just said to see the tests below since it's explained there.

PR #198, opened 2018-06-13
Ignore XDS and Text packets

squarebracket:
I noticed that XDS and Text packets aren't being ignored, and those may interfere with captions.

To ignore the packets, I do similar to what's done before any active channel is set -- I set the `activeCea608Channel` for the field to `null` so that packets are dropped.

Note: I've never seen Text packets in real life, so if anyone has any sources, please share!


gkatsev:
Should be good to go once it's rebased.

squarebracket:
Rebased to master, and added some comments

PR #201, opened 2018-07-19
docs: add some documentation to explain existing streams

ldayananda:


lib/codecs/h264.js
```diff
@@ -16,6 +16,13 @@ NalByteStream = function() {
     buffer;
   NalByteStream.prototype.init.call(this);
 
+  /*
+   * Scans a byte stream and triggers a data event with the NAL units found.
+   * @param {Object} data Event received from H264Stream
```
gesinger: This may not always come from the H264Stream, it's just the only source right now. It may be a bit "dangerous" to comment about a specifically named component from elsewhere, even though they're coupled pretty tightly right now, because the comment may become out-of-date without someone realizing it's there.

It may also be better to not consider the event either, as that's an upstream concern around the stream concept.

lib/codecs/h264.js
```diff
@@ -149,6 +156,17 @@ H264Stream = function() {
   H264Stream.prototype.init.call(this);
   self = this;
 
+  /*
+   * Pushes a packet from an Elementary Stream onto the NalByteStream
```
gesinger: Same comment as above around introducing Elementary Stream here (and for comments below).

lib/m2ts/m2ts.js
```diff
@@ -81,6 +84,9 @@ TransportPacketStream = function() {
     }
   };
 
+  /**
+   * Passes identified M2TS packets to the TransportParseStream to be parsed
```
gesinger: Same re: TransportParseStream

lib/codecs/h264.js
ldayananda: I think while NalByteStream is enclosed by H264Stream, it's highly unlikely that we would get other kinds of data. In fact, within this method we assume that we're dealing with a specific format(and box structure). 

To clarify further why I chose to mention H264Stream explicitly: we currently have zero indications that NalByteStream is packaged within H264Stream and they are coupled together. Without removing the Stream concept entirely I don't think we can stay away from referring to specific Streams as sources as it is not clear to an unfamiliar reader.

Personally, when I was new to the project, I had no clue what was going on here

>It may also be better to not consider the event either, as that's an upstream concern around the stream concept.

what do you mean by not consider?

lib/codecs/h264.js
ldayananda: This one I can agree can be more generic since it might actually be used in a different order

lib/m2ts/m2ts.js
ldayananda: I think the comment for NalByteStream would be applicable here too since TransportParseStream is contained within TransportPacketStream 🤔

PR #204, opened 2018-07-27
Create frame-utils module

gesinger:


lib/mp4/frame-utils.js
```diff
@@ -0,0 +1,242 @@
...
+  gops.byteLength += currentGop.byteLength;
+  gops.nalCount += currentGop.nalCount;
+  gops.duration += currentGop.duration;
+
+  // push the final Gop
+  gops.push(currentGop);
+  return gops;
+};
+
+var extendFirstKeyFrame = function(gops) {
```
ldayananda: Would be cool to add a contract/description of this method's purpose

lib/mp4/frame-utils.js
```diff
@@ -0,0 +1,242 @@
...
+      dependsOn: 1,
+      isDependedOn: 0,
+      hasRedundancy: 0,
+      degradationPriority: 0,
+      isNonSyncSample: 1
+    }
+  };
+};
+
+var sampleForFrame = function(frame, dataOffset) {
```
ldayananda: A comment here too would be awesome

ldayananda:
Does look like you need to rebase on master again though @gesinger

PR #206, opened 2018-07-31
Fix links in test content doc

gkatsev:


docs/test-content.md
```diff
@@ -38,14 +38,14 @@
 
 Then do *either* of the following:
 
-- Use [Bento4](#bento4) to split the file into an init segment and a fmp4 media segments:
+- Use [Bento4][bento4] to split the file into an init segment and a fmp4 media segments:
```
gesinger: Technically I don't think the bento4 links go anywhere.

gkatsev: they will point at `[bento4]: https://www.bento4.com/documentation/`

gesinger: You're right, I was actually clicking the one on line 31

gkatsev: oh, I guess I missed that one, let me update it too

PR #213, opened 2018-08-10
fix(mp4/Transmuxer): respect keepOriginalTimestamps for captions and ID3 in mp4

ldayananda:
There is a `keepOriginalTimestamps` flag which is intended to avoid adjusting timestamps when transmuxing from TS -> MP4. However, we do not respect this flag when returning captions and ID3 tags parsed out of the video data, leading to the captions and ID3 cues being out-of-sync with the video data.

Releasing this might be tricky as users may be depending on this behavior. A couple of ways we could do this:
- release as a major version, with captions and ID3 cues respecting the flag
- make a new flag to avoid adjusting the timing for captions and ID3 cues specifically

The downsides of the first is that any users relying on the current behavior with `keepOriginalTimestamps` would have to adjust caption times to start at zero if the TS segments do not start at zero.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -716,6 +716,9 @@ VideoSegmentStream.prototype = new Stream();
  * A Stream that can combine multiple streams (ie. audio & video)
  * into a single output segment for MSE. Also supports audio-only
  * and video-only streams.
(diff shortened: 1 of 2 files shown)


ldayananda:
Note: We have decided to release this as a new major version

PR #228, opened 2018-10-24
Support for Partial Appends

gesinger:
Includes work by: @mjneil @ldayananda @gesinger

This mux.js PR adds support for a partial data pipeline in mux.js. This is useful for features like LHLS (low latency HLS). It also adds support for emitting `data`, `caption`, `id3Frame`, `trackinfo`, and `timingInfo` as individual events, instead of waiting until a full segment has been processed. This provides a better mechanism for users to consume data from the partial pipeline, as well as write simpler handlers for processed data.

Although there are a good deal of changes, they exist either in the separate pipeline (the `partial` directory), or are add-ons to the existing transmuxer pipeline, thus, should all be backwards compatible.

`reset` and `endTimeline` were added to the primary transmuxer pipeline to allow for clearing any partially handled data, and to mark the end of a timeline (and clear out similar information).

`partialFlush` (which emits `partialdone`) was added for the partial pipeline, so that any completed data can be emitted before the next append is made.

A few TODOs exist that will have to be re-examined, and more tests should be added, but putting up this PR for some initial feedback.


lib/mp4/transmuxer.js
```diff
@@ -763,7 +788,27 @@ CoalesceStream.prototype.flush = function(flushSource) {
...
 
   // Emit the built segment
+  // We include captions and ID3 tags for backwards compatibility,
+  // ideally we should send only video and audio in the data event
   this.trigger('data', event);
+  // Emit each caption to the outside world
+  // Ideally, this would happen immediately on parsing captions,
+  // but we need to ensure that video data is sent back first
+  // so that caption timing can be adjusted to match video timing
+  for (i = 0; i < event.captions.length; i++) {
```
mjneil: Why do we need to send them out as individual events instead of a single event with just all the captions?  Same question for ID3 tags

lib/partial/audio-segment-stream.js
```diff
@@ -0,0 +1,164 @@
...
+    // throwing away hundreds of media segment fragments
+    boxes = new Uint8Array(moof.byteLength + mdat.byteLength);
+
+    boxes.set(moof);
+    boxes.set(mdat, moof.byteLength);
+
+    trackInfo.clearDtsInfo(track);
+
+    if (segmentStartDts === null) {
+      segmentStartDts = frames[0].dts;
```
mjneil: Feel free to ignore. Style choice `segmentStartDts = segmentEndDts = frames[0].dts;`

lib/partial/transmuxer.js
```diff
@@ -0,0 +1,375 @@
...
+      packet: new m2ts.TransportPacketStream(),
+      parse: new m2ts.TransportParseStream(),
+      elementary: new m2ts.ElementaryStream(),
+      videoRollover: new m2ts.TimestampRolloverStream('video'),
+      audioRollover: new m2ts.TimestampRolloverStream('audio'),
+      adts: new codecs.Adts(),
+      h264: new codecs.h264.H264Stream(),
+      captionStream: new m2ts.CaptionStream(),
+      metadataStream: new m2ts.MetadataStream(),
+      timedMetadataTimestampRolloverStream:
```
mjneil: Bit of a mouthful compared to the rest of the stream properties. Maybe `timedMetadataRollover` to match the audio and video versions

lib/partial/transmuxer.js
```diff
@@ -0,0 +1,375 @@
...
+        pipeline.tracks[data.tracks[i].type] = data.tracks[i];
+      }
+    }
+
+    if (pipeline.tracks.video && !pipeline.videoSegmentStream) {
+      pipeline.videoSegmentStream = new VideoSegmentStream(pipeline.tracks.video, options);
+
+      pipeline.videoSegmentStream.on('timelineStartInfo', function(timelineStartInfo) {
+        if (pipeline.tracks.audio) {
+          pipeline.audioSegmentStream.setEarliestDts(timelineStartInfo);
```
mjneil: The non-partial transmuxer sets `timelineStartInfo` on the audio track here which you are no longer doing. 
There is the comment 
```
          // When video emits timelineStartInfo data after a flush, we forward that
          // info to the AudioSegmentStream, if it exists, because video timeline
          // data takes precedence.
```

Is this no longer the case? are `timelineStartInfo` now being kept completely separate?

lib/partial/transmuxer.js
```diff
@@ -0,0 +1,375 @@
...
+  pipeline.metadataStream.on('timestamp', function(frame) {
+    pipeline.aacStream.setTimestamp(frame.timeStamp);
+  });
+
+  pipeline.aacStream.on('data', function(data) {
+    if (data.type !== 'timed-metadata' || pipeline.audioSegmentStream) {
+      return;
+    }
+
+    var audioTrack = {
```
mjneil: You should be using `pipeline.tracks.audio` here instead of creating a new variable. This will cause the `AudioSegmentStream` to be using a different track object than the transmuxer uses in `setBaseMediaDecodeTime`

lib/mp4/transmuxer.js
ldayananda: This matches the behavior of the partialTransmuxer which will get one caption/ID3 as they are encountered

lib/codecs/index.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  adts: require('./adts'),
+  Adts: require('./adts'),
```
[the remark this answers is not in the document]
gesinger: The linter didn't like it: "A constructor name should not start with a lowercase letter"

lib/m2ts/m2ts.js
```diff
@@ -301,6 +309,7 @@ ElementaryStream = function() {
       data: [],
       size: 0
     },
+    programMapTable,
```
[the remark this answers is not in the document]
gesinger: I think that was one of your commits: https://github.com/videojs/mux.js/pull/228/commits/d736b9c647c8bc160b2f96a0490899bbfd0832c7

lib/partial/transmuxer.js
gesinger: Nice catch.

In the non-partial transmuxer we only seem to use the `timelineStartInfo` in the audio track for `setEarliestDts`, where the earliest allowed dts would be the dts passed in as an argument minus the `timelineStartInfo.baseMediaDecodeTime`. Technically we can probably simplify the logic over in the non-partial transmuxer by removing that and doing the calculation before passing in the earliest dts.

In the partial transmuxer, things are a bit different, so we'll have to play with the logic a bit. `timelineStartInfo` is now triggered in the video segment stream before the first recalculation of `baseMediaDecodeTime`, whereas in the non-partial it was triggered after, thus, could be a different value. For the case of keeping original timestamps, we may be OK, but the logic may have to be changed around a bit.

gkatsev:
We should verify TBA work doesn't require anything in this PR. If there is, we may want to make a separate PR for it.
This likely doesn't affect regular HLS as it is a separate pipeline. It just requires some review and testing, in which case it can be merged in and issues can be fixed forward.
Make sure that captions work.

brandonocasey:
See #264

Issue #223, opened 2018-10-15
Caption-parser.js has an error in lastMatchedSample.pts


nicklive:
Hello!
I'm trying to create an hls stream with some sample file and have an error `Uncaught TypeError: Cannot read property 'pts' of undefined`

My environment:
> video.js: 7.2.2
> @videojs/http-streaming: 1.2.4
> mux.js: 4.5.1

<details>
 <summary>full log</summary>

``` 
Uncaught TypeError: Cannot read property 'pts' of undefined
    at findSeiNals (caption-parser.js:94)
    at eval (caption-parser.js:201)
    at Array.forEach (<anonymous>)
    at parseCaptionNals (caption-parser.js:183)
    at parseEmbeddedCaptions (caption-parser.js:237)
    at CaptionParser.parse (caption-parser.js:340)
    at eval (video.es.js:12759)
    at Object.eval [as callback] (video.es.js:10398)
    at cbOnce (index.js:63)
    at XMLHttpRequest.loadFunc (index.js:131)
findSeiNals @ caption-parser.js:94
(anonymous) @ caption-parser.js:201
parseCaptionNals @ caption-parser.js:183
parseEmbeddedCaptions @ caption-parser.js:237
CaptionParser.parse @ caption-parser.js:340
(anonymous) @ video.es.js:12759
(anonymous) @ video.es.js:10398
cbOnce @ index.js:63
loadFunc @ index.js:131
load (async)
_createXHR @ index.js:172
createXHR @ index.js:51
XhrFunction @ video.es.js:10392
mediaSegmentRequest @ video.es.js:12845
loadSegment_ @ video.es.js:13278
fillBuffer_ @ video.es.js:13194
monitorBufferTick_ @ video.es.js:13178
setTimeout (async)
monitorBuffer_ @ video.es.js:13173
load @ video.es.js:13095
(anonymous) @ video.es.js:14155
data.dispatcher @ video.es.js:798
trigger @ video.es.js:840
EventTarget.trigger @ video.es.js:1062
haveMetadata @ video.es.js:10143
(anonymous) @ video.es.js:10135
(anonymous) @ video.es.js:10398
cbOnce @ index.js:63
loadFunc @ index.js:131
load (async)
_createXHR @ index.js:172
createXHR @ index.js:51
XhrFunction @ video.es.js:10392
(anonymous) @ video.es.js:10134
data.dispatcher @ video.es.js:798
trigger @ video.es.js:840
[...]

PR #235, opened 2018-12-06
fix(caption-parser): do not fail when first sei sample is null

sgrogov:
By following up on <a href="https://github.com/videojs/mux.js/issues/223">that discussion</a>, maybe this small fix will be enough?

### DIFF
--- a/lib/mp4/caption-parser.js
+++ b/lib/mp4/caption-parser.js
@@ -89,11 +89,13 @@ var findSeiNals = function(avcStream, samples, trackId) {
         seiNal.pts = matchingSample.pts;
         seiNal.dts = matchingSample.dts;
         lastMatchedSample = matchingSample;
-      } else {
+      } else if (lastMatchedSample !== undefined) {
         // If a matching sample cannot be found, use the last
         // sample's values as they should be as close as possible
         seiNal.pts = lastMatchedSample.pts;
         seiNal.dts = lastMatchedSample.dts;
+      } else {
+        break;
       }
 
       result.push(seiNal);


nicklive:
Hi, guys! Any plans for merging this pr?

gkatsev:
@nicklive hey, do you have any test content you can share with us to test this?

carloscarnero:
I've observed [the same bug](https://github.com/videojs/mux.js/issues/223) in a couple of streams I have (at some point the player stops and the loading spinner begins). Manually modifying the source code as in this PR fixes this particular problem.

@gkatsev What would be a good method of sharing content? Currently, I can't publish it over HTTP but I could make it available to you by other means.

gkatsev:
If you can attach the file here to github that works, if you need to give it to me privately, feel free to either send it to me on [our slack](http://slack.videojs.com) or via email.

carloscarnero:
> If you can attach the file here to github that works, if you need to give it to me privately, feel free to either send it to me on [our slack](http://slack.videojs.com) or via email.

@gkatsev [Attached](https://github.com/videojs/mux.js/files/3023323/screensaver.zip) you can find a (hopefully) reduced test case scenario. Peeking on the console, it can be seen that the error manifests around 18 seconds, and playback stops at 1:06 (the 11th segment).

When using other players, the playback completes successfully.

gkatsev:
Thanks! I can see it failing and working with this fix. I wonder if we still want to push the seiNal unit into the results, even without the pts/dts values. @squarebracket thoughts?

Also, not sure what's up with travis.

squarebracket:
I can't see any reason that _not_ pushing an empty NAL would cause a problem, since the caption parser will disregard any non-caption packets before it does any processing on them. Might want to get @ldayananda to chime in here, though, since she wrote this particular bit of code.

ldayananda:
I think this change is hiding another potential issue. It looks like a matching sample is not found because the start of the SEI NAL in seg11 in the provided example starts at a point(`189975`) outside the samples in the stream(samples end at `189971`) so this call returns null

https://github.com/videojs/mux.js/blob/009f769128c82932949c99508dea5d2d0a9f4bd2/lib/mp4/caption-parser.js#L78

carloscarnero:
@ldayananda Does that means that there's an actual possibility that the transcoding/packaging process that generated those artifacts is defective? This is worrying for me since that process is completely controlled by me via custom ffmpeg pipelines; so that means that I'd need to revise them.

OTOH, neither hls.js nor other players (Bitmovin, JWPlayer, et al.)

gkatsev:
Closing in favor of #321.

PR #239, opened 2018-12-17
test: use standard browsers

forbesjo:
The build seems to be failing with headless Chrome, let's switch back to the standard browsers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ addons:
   firefox: latest
 before_install:
   - npm install -g greenkeeper-lockfile@1
-before_script: greenkeeper-lockfile-update
+before_script:
+  - greenkeeper-lockfile-update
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,6 +6,9 @@ addons:
   firefox: latest
 before_install:
   - npm install -g greenkeeper-lockfile@1
-before_script: greenkeeper-lockfile-update
+before_script:
+  - greenkeeper-lockfile-update
+  - export CHROME_BIN=/usr/bin/google-chrome
```
BrandonOCasey: was it actually failing because these entries were not in the config? or is it the headless browsers?

forbesjo: The Headless Chrome test were hanging on my local test run so I assume that there's something that changed with a Chrome update and one of our tests isn't completing.

BrandonOCasey: ahh I see

PR #243, opened 2019-01-16
Base videoSegmentTimingInfo on the baseMediaDecodeTime, rather than s…

gesinger:
…egment PTS and DTS values

* In addition, for reference and debugging, pass along the
baseMediaDecodeTime used

### DIFF
--- a/test/transmuxer.test.js
+++ b/test/transmuxer.test.js
@@ -1940,7 +1940,13 @@ QUnit.test('do not subtract the first frame\'s compositionTimeOffset from baseMe
 
 QUnit.test('video segment stream triggers segmentTimingInfo with timing info',
 function() {
-  var segmentTimingInfoArr = [];
+  var
+    segmentTimingInfoArr = [],
+    baseMediaDecodeTime = 40,
+    startingDts = 50,
+    startingPts = 60,
+    lastFrameStartDts = 150,
+    lastFrameStartPts = 160;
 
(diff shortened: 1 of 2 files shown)


test/transmuxer.test.js
```diff
@@ -1967,22 +1973,32 @@ function() {
...
+    baseMediaDecodeTime,
+    segmentTimingInfoArr[0].baseMediaDecodeTime,
+    'set baseMediaDecodeTime'
+  );
   QUnit.deepEqual(segmentTimingInfoArr[0], {
     start: {
-      dts: 50,
-      pts: 60
+      dts: baseMediaDecodeTime,
+      pts: baseMediaDecodeTime + startingPts - startingDts
```
ldayananda: I would just calculate out the value here in the expected result as we don't want the tests to repeat the code

test/transmuxer.test.js
```diff
@@ -2425,25 +2441,28 @@ function() {
...
       firstFrame.pts,
       lastFrame.dts + lastFrame.duration,
       lastFrame.pts + lastFrame.duration,
       prependedContentDuration
     ), {
       start: {
-        dts: 12,
-        pts: 14
+        dts: baseMediaDecodeTime,
+        pts: baseMediaDecodeTime + firstFrame.pts - firstFrame.dts
```
ldayananda: same for here

ldayananda:
should add a test where pts < dts for a gop since that's a thing that can happen

PR #247, opened 2019-02-09
Parse captions from tracks with trackId of zero

tylerdaines:
Currently, if the `trackId` is zero, the CaptionParser thinks it hasn't found an init segment yet and stores the segment for later inspection.  This PR changes the `trackId` checks in CaptionParser to allow for a `trackId` of zero, and thus the captions can then be parsed.

### DIFF
--- a/lib/mp4/caption-parser.js
+++ b/lib/mp4/caption-parser.js
@@ -231,7 +231,7 @@ var parseCaptionNals = function(segment, videoTrackId) {
 var parseEmbeddedCaptions = function(segment, trackId, timescale) {
   var seiNals;
 
-  if (!trackId) {
+  if (trackId === null) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


squarebracket:
Do you have a media where this is the case? I'm taking a look at the ISO-BMFF spec, and it seems that track ids cannot be zero. For instance, in the description for `track_ID` of the `tkhd`, it says:

> `track_ID` is an integer that uniquely identifies this track over the entire life‐time of this presentation. Track IDs are never re‐used and cannot be zero.

If you do have a media with a track id of 0, could you let us know what you used to transcode it?

tdaines:
Yeah, here is one example where the trackId is zero: https://content.uplynk.com/8aae79b846ca4cdeb88d9a06acbc5692.mpd.  We use a custom encoder that's been fixed only recently to not use 0.  Unfortunately we have 10's of thousands of assets that pre-date this change.

PR #248, opened 2019-02-18
Selection multiply fragments in one segment when findBox return 0

nicklive:
Hello, guys.
Recently I discovered an issue with byteranges and fragments in mp4.
This pull request fixes bug in `lib/mp4/probe.js`.
Brief description of the problem:
Construction `baseTime = baseTime || Infinity;` returns `Infinity` when  `baseTime === 0` so in can affect streams, that contains `tfhd` and function `findBox` return `0` as a valid number from mp4 file.

I add a test for this case as well

### DIFF
--- a/lib/mp4/probe.js
+++ b/lib/mp4/probe.js
@@ -142,8 +142,8 @@ startTime = function(timescale, fragment) {
   // determine the start times for each track
   baseTimes = [].concat.apply([], trafs.map(function(traf) {
     return findBox(traf, ['tfhd']).map(function(tfhd) {
-      var id, scale, baseTime;
-
+      var id, scale, baseTime, tfdt;
+      var baseTimeArray = [];
       // get the track id from the tfhd
       id = toUnsigned(tfhd[4] << 24 |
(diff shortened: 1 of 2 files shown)


lib/mp4/probe.js
```diff
@@ -153,24 +153,30 @@ startTime = function(timescale, fragment) {
...
-        if (version ===  1) {
-          result *= Math.pow(2, 32);
-          result += toUnsigned(tfdt[8]  << 24 |
-                               tfdt[9]  << 16 |
-                               tfdt[10] <<  8 |
-                               tfdt[11]);
-        }
-        return result;
-      })[0];
-      baseTime = baseTime || Infinity;
```
gkatsev: So, what you're saying is that baseTime can have a value of zero and we shouldn't set it to Infinity in this case which this line does.
I think we can simplify this change by just updating this line to something like
```js
if (typeof baseTime !== 'number') {
  baseTime = Infinity;
}
```
instead of any other change

gkatsev:
This is likely fixed via #356.

PR #262, opened 2019-04-22
chore: fix License inconsitencies

### LINKED ISSUE #259: License inconsistency
The tops of your source files say:

```js
/**
 * mux.js
 *
 * Copyright (c) 2016 Brightcove
 * All rights reserved.
 */
```

But your LICENSE file says:

```
Copyright 2015 Brightcove

Licensed under the Apache License, Version 2.0 (the "License");
you may not use this file except in compliance with the License.
You may obtain a copy of the License at

    http://www.apache.org/licenses/LICENSE-2.0
```

If this is intended to be licensed under Apache 2.0, can you please fix the tops of your source files to unambiguously state this instead of "all rights reserved"?  What is currently in LICENSE would make a good header in the source files.

And could you please update LICENSE to contain the [full text of the Apache 2.0 License](https://www.apache.org/licenses/LICENSE-2.0.txt), which contains the actual license terms in their full form?

Thanks!

gkatsev:
Update the LICENSE file to include the full text of the license, remove year from license, and add a header with a link to the license to each file.

Fixes #259

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,14 +1,202 @@
-Copyright 2015 Brightcove
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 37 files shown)


gkatsev:
Thanks @joeyparrish, it definitely counts, even if github doesn't say so :)

gkatsev:
And out in 5.1.2.

PR #267, opened 2019-05-13
chore: switch bundle-collapser to plugin usage

### LINKED ISSUE #173: "muxjs is not defined" error when using mux.min.js
With 'mux.js' everything is Ok. Looks like minified version doesn't expose 'window.muxjs' object

gkatsev:
Fixes #173, fixes #266


joeyparrish:
Confirmed that this works for me.

gkatsev:
Thanks @joeyparrish!

PR #270, opened 2019-05-17
feat: parse mp4 tracks

brandonocasey:
Parse all parts of mp4 tracks including codec information.


lib/mp4/probe.js
```diff
@@ -150,7 +150,7 @@ startTime = function(timescale, fragment) {
                       tfhd[6] <<  8 |
                       tfhd[7]);
       // assume a 90kHz clock if no timescale was specified
-      scale = timescale[id] || 90e3;
+      scale = (typeof timescale === 'object' ? timescale[id] : timescale) || 90e3;
```
ldayananda: are there times where timescale isn't an object?

lib/mp4/probe.js
```diff
@@ -221,10 +221,106 @@ getVideoTrackIds = function(init) {
...
+    if (stsd) {
+      var sampleDescriptions = stsd.subarray(8);
+      // gives the codec type string
+      track.codec = parseType(sampleDescriptions.subarray(4, 8));
+
+      // https://tools.ietf.org/html/rfc6381#section-3.3
+      if ((/^[a-z]vc[1-9]$/i).test(track.codec)) {
+        var codecConfig = sampleDescriptions.subarray(94);
+        // always avcc
+        // var type = parseType(codecConfig.subarray(4, 8));
```
ldayananda: what's happening here? Also why `subarray(94)`?

lib/mp4/probe.js
```diff
@@ -221,10 +221,106 @@ getVideoTrackIds = function(init) {
...
+        if (codecConfig.length > 20) {
+          track.codec += '.' + ('00' + codecConfig[19].toString(16)).slice(-2);
+          track.codec += '.' + (((codecConfig[20] >>> 2) & 0x3f).toString(16)).slice(-2);
+        }
+      } else {
+        // TODO: show a warning?
+      }
+    }
+
+    // TODO: should we parse timescale here?
```
ldayananda: I think we parse the timescales again later on the in the process using https://github.com/videojs/mux.js/blob/d8a5cd7a2a5fbe82b574dd47ea074d8f957987db/lib/mp4/probe.js#L86

lib/mp4/probe.js
brandonocasey: in the new `getTracks` function each track returns its own timescale

lib/mp4/probe.js
brandonocasey: If we find that we are dealing with an `avc1` codec we skip over the `avc1` box, as it doesn't contain anything relevant to parsing the codec parameters, and move to the next box. Which appears to be `avcC`. That box starts at `subarray(94)`. Perhaps we should use `findBox` here, so that it makes more sense.

lib/mp4/probe.js
brandonocasey: So the referenced function gets the timescale for all tracks. This function gets all the relevant information about every track including its timescale and stores it in a `track` object. I think that it makes more sense to do it that way. The TODO is mainly a question of wether we should track the `startTime` of each track separately or not, using their own timescale.

lib/mp4/probe.js
ldayananda: Personally, I like inputs to be of a consistent type since it's easier to reason about. That said, this is fine technically.

lib/mp4/probe.js
ldayananda: Yeah, I think a combination of using `findBox` and comments would help here

lib/mp4/probe.js
```diff
@@ -221,10 +221,106 @@ getVideoTrackIds = function(init) {
...
+          track.codec +=  ('00' + codecConfig[9].toString(16)).slice(-2);
+          // the byte containing the constraint_set flags
+          track.codec += ('00' + codecConfig[10].toString(16)).slice(-2);
+          // level idc
+          track.codec += ('00' + codecConfig[11].toString(16)).slice(-2);
+        }
+      } else if ((/^mp4[a,v]$/i).test(track.codec)) {
+        var codecConfig = sampleDescriptions.subarray(36);
+        // always esds
+        // var type = parseType(codecConfig.subarray(4, 8));
```
ldayananda: is this commented out for a reason?

lib/mp4/probe.js
ldayananda: hmm, considering we're bothering to get the timescales in this function, might as well also get the `startTime`s for each track as well, so we can get all the information back in one call

brandonocasey: we can only get the start time from an mp4 fragment, but we get the timescales from the mp4 init segment, so we have to get them at separate.

ldayananda: seems like you answered your TODO then 😄

lib/mp4/probe.js
brandonocasey: actually it seems like we want it to always be an object, and I just realized why. We are getting the start time via a fragment, which could have any of the track ids from the init segment.

brandonocasey: Made a new helper, rather than changing the type of timescale.

lib/mp4/probe.js
brandonocasey: We cannot, as we get `startTime` from mp4 fragments, but we get timescale from the regular mp4 init segment.

test/mp4-inspector.test.js
```diff
@@ -285,13 +287,164 @@ QUnit.test('can parse a moov', function() {
...
+                streamDescriptor: {
+                  decoderConfig: {
+                    avgBitrate: 0,
+                    bufferSize: 0,
+                    decoderConfigDescriptor: {
+                      audioObjectType: 0,
+                      channelConfiguration: 0,
+                      length: 0,
+                      samplingFrequencyIndex: 0,
+                      tag: 0
```
brandonocasey: I added another `trak` to the `sampleMoov`

test/utils/mp4-helpers.js
```diff
@@ -144,7 +144,124 @@ module.exports.sampleMoov =
...
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00, 0x00, 0x00, // box content
+                            0x00, 0x00,             // box content
+                            typeBytes('avcC'),      // codec profile type
```
ldayananda: this is awesome!

PR #272, opened 2019-06-01
 Basic support for FLAC-in-MP4 audio tracks

brion:
Allows for creating an MP4 stream with FLAC contents using the MP4 box generation functions or AudioSegmentStream.

Use codec string 'fLaC' and specify a 'metadata' field in the track info with the FLAC METADATA_BLOCK_HEADER fields to go into the dfLa box.

See https://github.com/xiph/flac/blob/master/doc/isoflac.txt for details of the MP4 embedding of FLAC audio.


squarebracket:
This is a cool feature, but we've got a lot on our plate right now so it might be a bit before we get a chance to review this.

brion:
No worries, no rush. :)

gkatsev:
@brion hey, do you have some test content readily available we can take a look at? Thanks.

brandonocasey:
@brion I tested out your pull request and I do see how this would help us generate flac-in-mp4 content. Since this doesn't seem to break mpeg ts transmuxing on our end I think we could merge this easily enough. It would be really helpful to get two things from you here so that we can get this in:

1. Do you have an example of how you are using the mp4-generator directly so that we can test it locally?
2. Can you add a unit test that runs through the mp4-in-flac generation code paths?

Issue #274, opened 2019-06-05
Missing pasp atom


emkman99:
Anamorphic (non-square pixels) video doesn't not play correctly in Windows 10 Chrome when hardware acceleration is turned on (https://github.com/videojs/http-streaming/issues/312). All other browsers/OS combinations seem to handle non-square pixels just fine. I've traced the issue back to a missing pasp atom in the mp4 generator (for comparison, hls.js includes the pasp atom [here](https://github.com/video-dev/hls.js/blob/master/src/remux/mp4-generator.js#L398)).

emkman99:
I just realized Chrome on Linux doesn't use hardware acceleration, so that's why it works there.

I'm not familiar with Chrome or Firefox source code, so someone correct me if I'm wrong. I found some areas that may explain the differences in how they handle anamorphic video.

- Chrome uses pasp only [(chrome mp4_stream_parser.cc#L498)](https://github.com/chromium/chromium/blob/9c5621cb5b4261d1187afaa722629bebde48bd40/media/formats/mp4/mp4_stream_parser.cc#L498), and ignores SPS data in the avcc atom [(chrome mp4_stream_parser.cc#L529)](https://github.com/chromium/chromium/blob/9c5621cb5b4261d1187afaa722629bebde48bd40/media/formats/mp4/mp4_stream_parser.cc#L529). There's also a [chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=843150) for this.

- Firefox uses the SPS data from the avcc atom and ignores the pasp [(firefox mp4parse#L1937)](https://github.com/mozilla/gecko-dev/blob/edce6c68ac3206a34ec382ce333074ad12c001c5/media/mp4parse-rust/mp4parse/src/lib.rs#L1937).

PR #275, opened 2019-06-05
fix: Added pasp atom to fix anamorphic video in win10 chrome (#274)

emkman99:
I built http-streaming 1.9.3 and videojs master (7.5.4) with this change. My test videos and the source from the original bug https://github.com/videojs/http-streaming/issues/312#issuecomment-447783343 play correctly.


squarebracket:
`pasp` is the pixel aspect ratio box, so that makes sense. Thanks for the investigation and fix!

A note from the spec:
> These are both optional; if present, they over‐ride the declarations (if any) in structures specific to the video codec, which structures should be examined if these boxes are absent. For maximum compatibility, these boxes should follow, not precede, any boxes defined in or required by derived specifications.

lib/codecs/h264.js
```diff
@@ -332,7 +332,7 @@ H264Stream = function() {
       picHeightInMapUnitsMinus1,
       frameMbsOnlyFlag,
       scalingListCount,
-      sarRatio,
+      sarRatio = [1, 1],
```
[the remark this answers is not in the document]
emkman99: hls.js always included the pasp atom so it had to set the default to 1:1. I just changed it around to only include the pasp atom when the aspect ratio is present. (All tests still pass in mux.js, http-streaming, and videojs.)

test/transmuxer.test.js
```diff
@@ -3412,7 +3412,7 @@ QUnit.test('generates a video init segment', function() {
   QUnit.ok(segments[0].info, 'video info is alongside video segments/bytes');
 
   mp4VideoProperties.forEach(function(prop) {
-    QUnit.ok(segments[0].info[prop], 'video info has ' + prop);
+    QUnit.ok(prop in segments[0].info, 'video info has ' + prop);
```
emkman99: This was the easiest way to "fix" this test when sarRatio is undefined. Is this acceptable?

squarebracket: I don't know for sure, but I'd imagine that this checks to make sure the MP4 init segment has all the properties passed through from the PES. So this change would mean the test is no longer testing for what it should.

Could you add the SAR data to the SPS data?

lib/mp4/mp4-generator.js
```diff
@@ -445,23 +465,36 @@ stbl = function(track) {
...
+          (vSpacing & 0xFF000000) >> 24,
+          (vSpacing & 0xFF0000) >> 16,
+          (vSpacing & 0xFF00) >> 8,
+          vSpacing & 0xFF
+        ]));
+
+      result = box(types.avc1, avc1Data,
+        avcCBox,
+        btrtBox,
+        paspBox
```
squarebracket: Might be cleaner to push the boxes onto an array and then do a `box.apply(null, boxes)` instead of the if/else.

test/transmuxer.test.js
emkman99: > I don't know for sure, but I'd imagine that this checks to make sure the MP4 init segment has all the properties passed through from the PES. So this change would mean the test is no longer testing for what it should.

mp4VideoProperties are from the SPS, not the PES header (if that's what you mean). I changed it back to how it was for every other property. I'm not sure how else to do it. (https://github.com/videojs/mux.js/pull/275/commits/2c2776277b356db5c177cadfeecf2581450c7854)

> Could you add the SAR data to the SPS data?

I'm not sure what you mean by this?

woellij:
looking forward to this PR getting merged. causing a lot of issues in video.js

woellij:
please more this

SebKei:
Please merge asap. Makes lots of trouble with IE11.

emkman99:
> Please merge asap. Makes lots of trouble with IE11.

I cannot reproduce this issue in win10 IE11.

SebKei:
> 
> 
> > Please merge asap. Makes lots of trouble with IE11.
> 
> I cannot reproduce this issue in win10 IE11.

Sorry ... too much open issues in parallel.
Of course I meant Chrome.
Exceptionally this time it's not IE causing trouble. ;)

marcelbeumer:
Would be great to have this merged, same issue here!

SebKei:
Thank you @emkman99 !

PR #276, opened 2019-06-19
Feat: change remux after initial creation

brandonocasey:
For fmp4 playlists with ts segment discontinuities the track ids of the fmp4 files and the ts segments that we transmux to fmp4 must match up.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -257,7 +257,8 @@ VideoSegmentStream = function(track, options) {
       }, this);
     }
 
-    if (nalUnit.nalUnitType === 'pic_parameter_set_rbsp' && !pps) {
+    if (nalUnit.nalUnitType === 'pic_parameter_set_rbsp' &&
+        !pps) {
       pps = nalUnit.data;
       track.pps = [nalUnit.data];
     }
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -21,8 +21,8 @@ var AdtsStream = require('../codecs/adts.js');
 var H264Stream = require('../codecs/h264').H264Stream;
 var AacStream = require('../aac');
 var isLikelyAacData = require('../aac/utils').isLikelyAacData;
+var ONE_SECOND_IN_TS = require('../utils/clock').ONE_SECOND_IN_TS;
 
-var ONE_SECOND_IN_TS = 90000; // 90kHz clock
 
 // constants
 var AUDIO_PROPERTIES = [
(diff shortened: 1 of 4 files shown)


lib/partial/transmuxer.js
```diff
@@ -352,6 +352,15 @@ var Transmuxer = function(options) {
     }
   };
 
+  this.setRemux = function(val) {
```
brandonocasey: currently this won't do anything as we don't have a `coalesceStream` for the partial transmuxer. This also means fmp4 and ts content will not work with the partial transmuxer right now.

ldayananda: I believe it was intentional that there is no coalesceStream for the partial transmuxer in order to reduce the time to return data

lib/mp4/transmuxer.js
```diff
@@ -709,14 +708,21 @@ CoalesceStream = function(options, metadataStream) {
     // important information required for the construction of
     // the final segment
     this.pendingTracks.push(output.track);
-    this.pendingBoxes.push(output.boxes);
     this.pendingBytes += output.boxes.byteLength;
 
+    // TODO: is there an issue for this against chrome?
```
brandonocasey: This is what caused the test change, and a lot of the headache I was seeing getting this to work

lib/mp4/transmuxer.js
```diff
@@ -21,8 +21,8 @@ var AdtsStream = require('../codecs/adts.js');
 var H264Stream = require('../codecs/h264').H264Stream;
 var AacStream = require('../aac');
 var isLikelyAacData = require('../aac/utils').isLikelyAacData;
+var ONE_SECOND_IN_TS = require('../utils/clock').ONE_SECOND_IN_TS;
```
brandonocasey: This isn't a required change but we use it this way in most other files. I think I missed this during the original rebase.

lib/partial/audio-segment-stream.js
```diff
@@ -4,6 +4,7 @@ var Stream = require('../utils/stream.js');
 var mp4 = require('../mp4/mp4-generator.js');
 var audioFrameUtils = require('../mp4/audio-frame-utils');
 var trackInfo = require('../mp4/track-decode-info.js');
+var ONE_SECOND_IN_TS = require('../utils/clock').ONE_SECOND_IN_TS;
```
brandonocasey: This isn't a required change but we use it this way in most other files. I think I missed this during the original rebase.

test/transmuxer.test.js
```diff
@@ -3738,7 +3738,7 @@ validateTrackFragment = function(track, segment, metadata, type) {
 QUnit.test('parses an example mp2t file and generates combined media segments', function() {
   var
     segments = [],
-    i, j, boxes, mfhd, trackType = 'video', trackId = 256, baseOffset = 0, initSegment;
+    i, j, boxes, mfhd, trackType = 'audio', trackId = 257, baseOffset = 0, initSegment;
```
brandonocasey: Since we now always output video first, this test had to be changed.

lib/mp4/transmuxer.js
```diff
@@ -1119,6 +1129,16 @@ Transmuxer = function(options) {
     }
   };
 
+  this.setRemux = function(val) {
```
brandonocasey: This allows us to set the `remux` option after the transmuxer is created.

PR #277, opened 2019-06-21
fix: Fire trackinfo for audio only transmuxers

brandonocasey:
Right now we only fire `trackinfo` for `tsPipelines`, but  we should fire it for `aacPipelines` as well.

### DIFF
--- a/lib/partial/transmuxer.js
+++ b/lib/partial/transmuxer.js
@@ -165,11 +165,7 @@ var aacPipeline = function(options) {
     pipeline = {
     type: 'aac',
     tracks: {
-      audio: {
-        timelineStartInfo: {
-          baseMediaDecodeTime: options.baseMediaDecodeTime
-        }
-      }
+      audio: null
     },
     metadataStream: new m2ts.MetadataStream(),
     aacStream: new AacStream(),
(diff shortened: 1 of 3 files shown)


lib/partial/transmuxer.js
```diff
@@ -197,9 +193,9 @@ var aacPipeline = function(options) {
       return;
     }
 
-    pipeline.tracks.audio = {
+    pipeline.tracks.audio = pipeline.tracks.audio || {
```
brandonocasey: Made this more like the non-partial transmuxer.

lib/partial/transmuxer.js
```diff
@@ -208,19 +204,13 @@ var aacPipeline = function(options) {
     // hook up the audio segment stream to the first track with aac data
     pipeline.audioSegmentStream = new AudioSegmentStream(pipeline.tracks.audio, options);
 
-    pipeline.audioSegmentStream.on('timingInfo',
```
brandonocasey: We were firing audioTimingInfo twice, woops

lib/partial/transmuxer.js
```diff
@@ -208,19 +204,13 @@ var aacPipeline = function(options) {
...
-      pipeline.trigger.bind(pipeline, 'audioTimingInfo'));
-
-    // Set up the final part of the audio pipeline
-    pipeline.adtsStream
```
brandonocasey: Move the actual pipe below listener definitions just in case.

lib/partial/transmuxer.js
```diff
@@ -266,6 +264,7 @@ var Transmuxer = function(options) {
     hasFlushed = true;
 
   Transmuxer.prototype.init.call(this);
+  options.baseMediaDecodeTime = options.baseMediaDecodeTime || 0;
```
brandonocasey: Without this baseMediaDecodeTime defaults to undefined, which is weird.

PR #279, opened 2019-06-26
fix: If there isn't any video OR audio track data, just emit done

misteroneill:
There are scenarios where we parsed timed-metadata but did NOT find any video or audio data. For timing reasons, we can't map the timed-metadata to a proper cue-time until we have received PTS timing data from audio or video tracks. This change causes the transmuxer to hold onto any parsed timed-metadata info instead of emitting it, treating that timed-metadata like it came from the NEXT segment (which hopefully has content) so that we can do a proper mapping to generate the cues' startTime and endTime.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -777,88 +777,90 @@ CoalesceStream.prototype.flush = function(flushSource) {
     }, this);
   }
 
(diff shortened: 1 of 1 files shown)


misteroneill:
Definitely hide whitespace changes here.

PR #285, opened 2019-07-19
fix: Fix an issue where a rollover in only one audio or video track could cause the other to be dropped.

misteroneill:
### Problem
We saw an issue where some streams had a discontinuity and DTS rollover in one track (e.g. video), but not in another (e.g. audio). The effect of this was that the audio would be interpreted as being ~23 hours in the past and would be thrown out.

For live streams, this meant that the stream would abruptly halt and ultimately cause the player to time out. For VOD streams - at least, in the test case we have - this means audio would be dropped, but video would continue to play.

### Solution
The solution here is to use a single, shared `TimestampRolloverStream` for cases where this is possible.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -1103,6 +1103,9 @@ Transmuxer = function(options) {
       if (!options.keepOriginalTimestamps) {
         audioTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
       }
+      if (pipeline.audioTimestampRolloverStream) {
+        pipeline.audioTimestampRolloverStream.discontinuity();
+      }
       if (pipeline.timestampRolloverStream) {
         pipeline.timestampRolloverStream.discontinuity();
       }
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -1111,7 +1111,6 @@ Transmuxer = function(options) {
     if (videoTrack) {
       if (pipeline.videoSegmentStream) {
         pipeline.videoSegmentStream.gopCache_ = [];
-        pipeline.timestampRolloverStream.discontinuity();
       }
       videoTrack.timelineStartInfo.dts = undefined;
       videoTrack.timelineStartInfo.pts = undefined;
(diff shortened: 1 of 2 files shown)


lib/mp4/transmuxer.js
```diff
@@ -987,14 +985,12 @@ Transmuxer = function(options) {
     // !!THIS ORDER IS IMPORTANT!!
     // demux the streams
     pipeline.elementaryStream
-      .pipe(pipeline.videoTimestampRolloverStream)
+      .pipe(pipeline.timestampRolloverStream)
```
mjneil: This should be moved up to line 984 and then the `pipeline.elementaryStream` piping into each type here should be changed to `pipeline.timestampRolloverStream`

lib/mp4/transmuxer.js
```diff
@@ -1128,6 +1124,10 @@ Transmuxer = function(options) {
     if (pipeline.timedMetadataTimestampRolloverStream) {
```
gkatsev: this should be removed

lib/mp4/transmuxer.js
```diff
@@ -982,19 +980,17 @@ Transmuxer = function(options) {
     // disassemble MPEG2-TS packets into elementary streams
     pipeline.packetStream
       .pipe(pipeline.parseStream)
-      .pipe(pipeline.elementaryStream);
+      .pipe(pipeline.elementaryStream)
+      .pipe(pipeline.timestampRolloverStream);
 
     // !!THIS ORDER IS IMPORTANT!!
     // demux the streams
     pipeline.elementaryStream
```
mjneil: the three following `pipeline.elementaryStream` should be `pipeline.timestampRolloverStream` so its more clear how the pipeline is structured

lib/mp4/transmuxer.js
```diff
@@ -1114,7 +1111,7 @@ Transmuxer = function(options) {
     if (videoTrack) {
       if (pipeline.videoSegmentStream) {
         pipeline.videoSegmentStream.gopCache_ = [];
-        pipeline.videoTimestampRolloverStream.discontinuity();
+        pipeline.timestampRolloverStream.discontinuity();
```
mjneil: its probably safe to delete this line, since there is a check for the rolloverstream at the end of this function

gkatsev:
Just verified locally as well.

PR #288, opened 2019-07-30
perf: make cone of silence lazy

brandonocasey:
Saves about ~10-20ms of script loading time in the browser in slow browsers/computers.

### DIFF
--- a/lib/data/silence.js
+++ b/lib/data/silence.js
@@ -23,19 +23,26 @@ var makeTable = function(metaTable) {
   }, {});
 };
 
-// Frames-of-silence to use for filling in missing AAC frames
-var coneOfSilence = {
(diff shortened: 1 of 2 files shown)


gkatsev:
We should find test content for this to make sure that the 10-20ms hit at usage time doesn't cause issues (unlikely but you never know).

brandonocasey:
It will probably be less than 10-20ms at that time too since startup is when the most stuff is going on

gkatsev:
That's true. Would be good to know if there is an impact, though, it is likely smaller.

brandonocasey:
just an update here, as this will probably be merged eventually. I tested this with one of the streams that this code originally fixed, and it is completely un-impacted by this change.

PR #293, opened 2019-08-15
fix: use metadataTsToSeconds for id3 as we did previously

brandonocasey:
Previous to TBA/LHLS we did code that equated to `metadataTsToSeconds` on `id3`s. In master we do `videoTsToSeconds`. The code itself seems to think that `metadataTsToSeconds` should be called as it passes all the parameters for it, and it fixes an issues where cues are not firing at the correct time, since they are trying to use video time and not metadata time.

This means we can close #292 (hopefully)

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -828,7 +828,7 @@ CoalesceStream.prototype.flush = function(flushSource) {
     // video timeline for the segment
     for (i = 0; i < this.pendingMetadata.length; i++) {
       id3 = this.pendingMetadata[i];
-      id3.cueTime = clock.videoTsToSeconds(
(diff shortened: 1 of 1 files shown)


lib/mp4/transmuxer.js
```diff
@@ -828,7 +828,7 @@ CoalesceStream.prototype.flush = function(flushSource) {
     // video timeline for the segment
     for (i = 0; i < this.pendingMetadata.length; i++) {
       id3 = this.pendingMetadata[i];
-      id3.cueTime = clock.videoTsToSeconds(
+      id3.cueTime = clock.metadataTsToSeconds(
```
brandonocasey: This seems like a typo to me as we are passing all the correct parameters, just calling the wrong function, woops

PR #300, opened 2019-09-10
fix: emit pts values instead of dts

brandonocasey:
See https://github.com/videojs/http-streaming/pull/624

It seems that most browsers use pts these days, and we get better results with it.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -193,8 +193,8 @@ AudioSegmentStream = function(track, options) {
     // frames. Leaving for now, but should be looked into.
     if (frames.length) {
       this.trigger('timingInfo', {
-        start: frames[0].dts,
-        end: frames[0].dts + (frames.length * frameDuration)
+        start: frames[0].pts,
+        end: frames[0].pts + (frames.length * frameDuration)
       });
     }
     this.trigger('data', {track: track, boxes: boxes});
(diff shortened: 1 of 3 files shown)


gkatsev:
@joeyparrish we've recently started getting more and more issues around video underflow, which stems around our usage of DTS here. From what we can tell, we used to use DTS here due to some bug in an old version of Firefox that used DTS timing rather than PTS timing but given that all browsers have moved to PTS we should do the same.
However, we want to make sure that Shaka doesn't break if we make this change and are considering whether this should be a patch/minor or maybe a major version update to mux.js.
Any input on this change and whether and how it'll affect Shaka would be appreciated.

joeyparrish:
@gkatsev, I think PTS is correct, and you should make the change.  I'll leave it to your discretion what the version number change should be.  We'll test the new version as soon as you can point me at a minified build that always does PTS, and if we need to adjust something in Shaka Player to match, we will.  If it's a trivial upgrade, we'll make it in the current v2.5.x branch, and if not, we'll make it in the next minor release, v2.6.0.

gkatsev:
Thanks @joeyparrish. We have a couple of other related changes that are in the pipeline, so, we may just do it all at the same time and do a major version update to just to be on the safe side.

joeyparrish:
Okay, sounds good.

gkatsev:
Let's make it a minor and if anyone complains, we can revert and make it a major

PR #301, opened 2019-09-17
feat: Add streamSelection options.

amotzte:
This option allow to get the final stream demuxed, either by stream type
or by pid.

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -114,14 +114,14 @@ TransportPacketStream.prototype = new Stream();
  * Accepts an MP2T TransportPacketStream and emits data events with parsed
  * forms of the individual transport stream packets.
  */
-TransportParseStream = function() {
+TransportParseStream = function(options) {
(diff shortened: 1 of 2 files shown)


gkatsev:
@amotzte can you provide some more information around the usecase for this? Thanks.

amotzte:
Hi @gkatsev currently mux.js takes only the first video pid and the first audio pid in the mpegts stream 
It is done in this code 

https://github.com/videojs/mux.js/blob/d31fa733571deb04a91bffde5e676e2f4127d32a/lib/m2ts/m2ts.js#L472-L481

As you can see, if an audio had already found, it will not go into the ```if``` and will not store the other audio pid.

This change allow 2 things
1. Select another audio pid (which is not the first one)
2. Select video only or audio to be transmuxed (basically leave doing demux without doing the mux again for fMP4)

gkatsev:
Thanks for clarifying. Makes sense, I hadn't even considered choosing a separate track.
Would you be able to add some tests for this?

PR #310, opened 2019-10-31
fix(mp4Inspector): read version 1 sidx boxes

### LINKED ISSUE #287: Incorrectly parsing SIDX box
Hi, there seems to be a bug with parsing the `SIDX` box from the mp4. I looked into the function that parses the SIDX and I believe this is where the bug exists.
https://github.com/videojs/mux.js/blob/b8df0af5f003b84267064789b3f0421d8f00f7cd/lib/tools/mp4-inspector.js#L390
The issue seems to be the way the  `earliestPresentationTime` and `firstOffset` are being parsed. In the `ISO/IEC 14496-12:2012` spec, in section `8.16.3.2`, the syntax dictates that the properties `earliestPresentationTime` and `firstOffset` are 64 bit values if the version is set to the value of `1`. They are 32 bit values if  the version is `0`. It looks like in your code, it assumes that those properties are always parsed as 32 bit values. I know it's not possible to represent a 64 bit integer in Javascript since it is limited to 53 bits. Our solution was to just parse 53 bits and move the index 8 bytes. We also log a warning message that only 53 bits of the value are being parsed.

ldayananda:
Should resolve #287.

### DIFF
--- a/lib/tools/parse-sidx.js
+++ b/lib/tools/parse-sidx.js
@@ -1,18 +1,34 @@
+var MAX_UINT32 = Math.pow(2, 32);
+
 var parseSidx = function(data) {
   var view = new DataView(data.buffer, data.byteOffset, data.byteLength),
-    result = {
-      version: data[0],
-      flags: new Uint8Array(data.subarray(1, 4)),
-      references: [],
-      referenceId: view.getUint32(4),
-      timescale: view.getUint32(8),
-      earliestPresentationTime: view.getUint32(12),
-      firstOffset: view.getUint32(16)
-    },
-    referenceCount = view.getUint16(22),
-    i;
+      result = {
+        version: data[0],
+        flags: new Uint8Array(data.subarray(1, 4)),
+        references: [],
+        referenceId: view.getUint32(4),
+        timescale: view.getUint32(8)
+      },
+    i = 12;
 
-  for (i = 24; referenceCount; i += 12, referenceCount--) {
+  if (result.version === 0) {
+    result.earliestPresentationTime = view.getUint32(i);
+    result.firstOffset = view.getUint32(i + 4);
+    i += 8;
+  } else {
+    // read 64 bits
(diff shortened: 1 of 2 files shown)


lib/tools/mp4-inspector.js
```diff
@@ -445,14 +445,53 @@ var
...
-      for (i = 24; referenceCount; i += 12, referenceCount--) {
+      if (result.version === 0) {
+        earliestPresentationTime = view.getUint32(i);
+        i += 4;
+        firstOffset = view.getUint32(i);
+        i += 4;
+      } else {
+        // read 64 bits
+        earliestPresentationTime = view.getUint32(i);
+        earliestPresentationTime *= Math.pow(2, 32);
```
amotzte: Regular integer in javascript is 52 bits. Might want to use [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) for 64 bits number.

gkatsev: Unfortunately, browser compatibility doesn't allow us to use BigInt yet. However, since we're working with Uint32 array views, we should still be able to access the full value.

brandonocasey (approved):

lib/tools/parse-sidx.js
```diff
@@ -1,18 +1,34 @@
...
+    i += 16;
+  }
+
+  i += 2; // reserved
+
+  var referenceCount = view.getUint16(i);
+
+  i += 2; // start of references
+
+  for (i; referenceCount; i += 12, referenceCount--) {
```
gesinger, line 31:
```suggestion
  for (; referenceCount > 0; i += 12, referenceCount--) {
```

gesinger (approved):
Looks good, but one style suggestion (though not required)

brandonocasey:
accidentally merged this into master. I cherry picked this into main and reverted the change in master.

brandonocasey:
Fixes https://github.com/videojs/http-streaming/issues/909

PR #313, opened 2019-11-15
perf: Single allocation of Uint8Array in trun.

amotzte:
This change improve performance by 3X (measured with node.js)

Instead of using array.concat when we iterate the frames (Which create
a new array + copy data from old to new for each iteration)
We are now creating a one time single array with the size of the
final array and filling it up.


gkatsev:
Thanks @amotzte!

PR #317, opened 2020-01-17
fix: always calculate the earliest allowed dts using baseMediaDecodeTime

brandonocasey:
## Problem 
`timelineStartInfo.baseMediaDecodeTime` is `0` when `keepOriginalTimestampOffsets` is `true`. This causes earliestAllowedDts to be much higher than it should be, which means that almost all audio is trimmed when seeking backwards for live stream videos. 

## Solution
Track `baseMediaDecodeTime` even though we don't intend to use it for most things when `keepOriginalTimestampOffsets` is `true`. That way we can use it to get the real timeline start dts.

## Testing
This fixes the stream mentioned in https://github.com/videojs/http-streaming/pull/624#issuecomment-544506799 that can be tested by turning on the `liveui` in the vhs test page and linking in this version of mux.js

### DIFF
--- a/lib/constants/audio-properties.js
+++ b/lib/constants/audio-properties.js
@@ -1,0 +1,10 @@
+// constants
+var AUDIO_PROPERTIES = [
+  'audioobjecttype',
+  'channelcount',
+  'samplerate',
+  'samplingfrequencyindex',
+  'samplesize'
+];
+
+module.exports = AUDIO_PROPERTIES;
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -132,7 +132,7 @@ AudioSegmentStream = function(track, options) {
   };
 
   this.setEarliestDts = function(earliestDts) {
-    earliestAllowedDts = earliestDts - track.timelineStartInfo.baseMediaDecodeTime;
+    earliestAllowedDts = earliestDts;
   };
 
   this.setVideoBaseMediaDecodeTime = function(baseMediaDecodeTime) {
@@ -1005,6 +1005,8 @@ Transmuxer = function(options) {
     pipeline.elementaryStream.on('data', function(data) {
       var i;
 
+      var baseMediaDecodeTime = !options.keepOriginalTimestamps ? self.baseMediaDecodeTime : 0;
+
       if (data.type === 'metadata') {
         i = data.tracks.length;
 
@@ -1012,10 +1014,10 @@ Transmuxer = function(options) {
         while (i--) {
           if (!videoTrack && data.tracks[i].type === 'video') {
             videoTrack = data.tracks[i];
-            videoTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
+            videoTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
           } else if (!audioTrack && data.tracks[i].type === 'audio') {
             audioTrack = data.tracks[i];
-            audioTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
+            audioTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
           }
         }
 
@@ -1034,7 +1036,7 @@ Transmuxer = function(options) {
               // very earliest DTS we have seen in video because Chrome will
               // interpret any video track with a baseMediaDecodeTime that is
               // non-zero as a gap.
-              pipeline.audioSegmentStream.setEarliestDts(timelineStartInfo.dts);
+              pipeline.audioSegmentStream.setEarliestDts(timelineStartInfo.dts - self.baseMediaDecodeTime);
             }
           });
 
@@ -1096,17 +1098,17 @@ Transmuxer = function(options) {
   this.setBaseMediaDecodeTime = function(baseMediaDecodeTime) {
     var pipeline = this.transmuxPipeline_;
 
+    this.baseMediaDecodeTime = baseMediaDecodeTime;
+
     if (!options.keepOriginalTimestamps) {
-      this.baseMediaDecodeTime = baseMediaDecodeTime;
+      baseMediaDecodeTime = 0;
     }
 
     if (audioTrack) {
       audioTrack.timelineStartInfo.dts = undefined;
       audioTrack.timelineStartInfo.pts = undefined;
       trackDecodeInfo.clearDtsInfo(audioTrack);
-      if (!options.keepOriginalTimestamps) {
-        audioTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
-      }
+      audioTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
       if (pipeline.audioTimestampRolloverStream) {
         pipeline.audioTimestampRolloverStream.discontinuity();
       }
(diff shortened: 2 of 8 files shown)


lib/mp4/transmuxer.js
```diff
@@ -132,7 +132,7 @@ AudioSegmentStream = function(track, options) {
   };
 
   this.setEarliestDts = function(earliestDts) {
-    earliestAllowedDts = earliestDts - track.timelineStartInfo.baseMediaDecodeTime;
+    earliestAllowedDts = earliestDts - videoBaseMediaDecodeTime;
```
brandonocasey: To me this seemed like a better solution than using the `timelinestartinfo.baseMediaDecodeTime` because this is directly related to the current video `baseMediaDecodeTime` and not the initial `baseMediaDecodeTime`. 

Said another way it is used for trimming audio before the video start. We want to use an accurate `baseMediaDecodeTime` regardless of wether `keepOriginalTimestampOffsets` is `true` or `false` and the only way to do that is to use `videoBaseMediaDecodeTime`.

test/transmuxer.test.js
```diff
@@ -3170,6 +3170,7 @@ QUnit.test('ensures baseMediaDecodeTime for audio is not negative', function() {
     events.push(event);
   });
   audioSegmentStream.track.timelineStartInfo.baseMediaDecodeTime = 10;
```
brandonocasey: kept this line around because this test actually verifies that this base media decode time is set on one of the boxes.

lib/mp4/transmuxer.js
gesinger: Because this is now using the `baseMediaDecodeTime` for a given segment, wouldn't this mean that any audio preceding video in the segment would be trimmed? Meaning that if a previous segment had an audio end shorter than video end, and the current segment fills that audio gap by having an audio start before video start, this would trim those audio frames and leave a gap in the audio buffer.

brandonocasey: You could be right hmmm, maybe this fix is going to be a bit more involved.

test/transmuxer.test.js
```diff
@@ -3170,7 +3170,7 @@ QUnit.test('ensures baseMediaDecodeTime for audio is not negative', function() {
...
-  audioSegmentStream.setEarliestDts(111);
+  audioSegmentStream.setEarliestDts(101);
```
brandonocasey: since baseMediaDecodeTime isn't used and this function is just a dumb setter now, we have to subtract ten so the tests stays mostly the same.

lib/mp4/transmuxer.js
```diff
@@ -1096,17 +1098,17 @@ Transmuxer = function(options) {
   this.setBaseMediaDecodeTime = function(baseMediaDecodeTime) {
     var pipeline = this.transmuxPipeline_;
 
+    this.baseMediaDecodeTime = baseMediaDecodeTime;
+
     if (!options.keepOriginalTimestamps) {
-      this.baseMediaDecodeTime = baseMediaDecodeTime;
+      baseMediaDecodeTime = 0;
```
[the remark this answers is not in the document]
brandonocasey: I think you are correct on both points here, going to push up a commit that removes setting timelineStartInfo.baseMediaDecodeTime which should remove the funny values that you also pointed out.

brandonocasey:
Good idea

brandonocasey:
Added tests for partial and full transmuxers I also pushed up a branch which should fail in master. As that branch only contains the test commit from this pull request.  https://travis-ci.org/videojs/mux.js/builds/646976496

lib/constants/audio-properties.js
```diff
@@ -0,0 +1,10 @@
+// constants
```
brandonocasey: Noticed that the partial transmuxer and the full transmuxer both use these. I can take this change out of this pull request if we want, but I kept it in for now as it was required at one point when I was adding tests for the partial transmuxer.

test/transmuxer.test.js
```diff
@@ -3408,6 +3408,416 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
...
+    earliestDts = 15000,
+    transmuxer = new Transmuxer({
+      keepOriginalTimestamps: false
+    });
+
+  transmuxer.on('data', function(segment) {
+    segments.push(segment);
+  });
+  transmuxer.on('done', function(segment) {
+    if (!segmentLengthOnDone) {
```
gesinger: Is this checked in case there are multiple dones?

test/transmuxer.test.js
```diff
@@ -3408,6 +3408,416 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
...
+    0x06, 0xb6, 0xc2, 0xb5,
+    0xef, 0x7c, 0x04
+  ], false, earliestDts)));
+  transmuxer.push(packetize(videoPes([
+      0x05, 0x01 // slice_layer_without_partitioning_rbsp_idr
+  ], true, earliestDts)));
+  transmuxer.flush();
+
+  QUnit.equal(segments.length, 1, 'generated a combined segment');
+  QUnit.equal(segments[0].data.length, 295, 'trimmed audio frame');
```
gesinger: What is this being compared to? Might just need a comment somewhere.

test/transmuxer.test.js
```diff
@@ -3408,6 +3408,416 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
...
+    0x06, 0xb6, 0xc2, 0xb5,
+    0xef, 0x7c, 0x04
+  ], false, earliestDts)));
+  transmuxer.push(packetize(videoPes([
+      0x05, 0x01 // slice_layer_without_partitioning_rbsp_idr
+  ], true, earliestDts)));
+  transmuxer.flush();
+
+  QUnit.equal(segments.length, 1, 'generated a combined segment');
+  QUnit.equal(segments[0].data.length, 305, 'trimmed audio frame');
```
gesinger: I think this should read not "did not trim audio frames" (and for others)

test/transmuxer.test.js
brandonocasey: I am just going to remove the `segmentLengthOnDone` code as it is irrelevant to the test.

test/transmuxer.test.js
brandonocasey: I will add a comment

brandonocasey:
When updating the tests, and expanding the partial transmuxer tests. I noticed that #285 was never done for the partial transmuxer, this caused half of the tests that we are looping over to fail in the partial transmuxer.

lib/partial/transmuxer.js
```diff
@@ -317,9 +315,7 @@ var Transmuxer = function(options) {
   };
 
   this.setBaseMediaDecodeTime = function(baseMediaDecodeTime) {
-    if (!options.keepOriginalTimestamps) {
```
gkatsev: so basically always track BMDT regardless of this option which will either be a value we care about or zero, in which case it won't do anything?

lib/partial/transmuxer.js
```diff
@@ -27,38 +27,34 @@ var tsPipeline = function(options) {
       packet: new m2ts.TransportPacketStream(),
       parse: new m2ts.TransportParseStream(),
       elementary: new m2ts.ElementaryStream(),
-      videoRollover: new m2ts.TimestampRolloverStream('video'),
```
gkatsev: this ends up matching up our non-partial transmuxer, right?

lib/partial/transmuxer.js
brandonocasey: Exactly

lib/partial/transmuxer.js
brandonocasey: Yeah, this was the rollover fix pat did awhile ago, but it never made it to the partial transmuxer. Without this the tests, and certainly some streams, were failing because the dts time for audio was much higher (never rolled over) then the video

PR #319, opened 2020-02-10
test: add vhs test page for netlify testing

brandonocasey:
This allows us to test vhs with mux.js changes using netlify

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "dist/"
   ],
   "devDependencies": {
-    "@videojs/http-streaming": "^1.12.0",
     "browserify": "^12.0.1",
     "browserify-shim": "^3.8.12",
     "bundle-collapser": "^1.2.1",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "main": "./lib/index.js",
   "scripts": {
+    "netlify": "node scripts/netlify.js",
     "test": "npm run lint && karma start test/karma/karma.conf.js",
     "start": "npm-run-all -p watch serve",
     "serve": "node scripts/server.js",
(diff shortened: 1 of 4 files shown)


scripts/netlify.js
```diff
@@ -0,0 +1,10 @@
+const shell = require('shelljs');
+
+shell.exec('git clone https://github.com/videojs/http-streaming');
```
brandonocasey: didn't see a better way to do this since vhs needs to build mux.js into itself.

misteroneill: Git submodules! :trollface:

scripts/netlify.js
```diff
@@ -0,0 +1,10 @@
...
+shell.cd('http-streaming');
+shell.exec('npm ci');
+shell.exec('npm link ../');
+shell.exec('npm run netlify');
```
gkatsev: can we add a comment to what this is doing? Specifically, that's it's running the netlify build of VHS itself.

scripts/netlify.js
```diff
@@ -0,0 +1,10 @@
...
+shell.cp('-R', 'deploy',  '../');
```
gkatsev: if we wanted to, we could configure it to deploy `http-streaming/deploy` then we don't need these 3 lines.

package.json
```diff
@@ -59,6 +60,7 @@
     "dist/"
   ],
   "devDependencies": {
+    "@videojs/http-streaming": "^1.12.0",
```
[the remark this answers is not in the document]
brandonocasey: nope

scripts/netlify.js
brandonocasey: We could, but I think it's better to run these lines and keep netlify `deploy` in a predictable place for all of our projects.

PR #329, opened 2020-04-06
fix: Do not trim audio output when keepOriginalTimestamps is set

### LINKED ISSUE #328: Fail to append buffer for multiplexed content with discontinuity
Shaka Player uses mux.js to convert TS to MP4. For multiplexed content with discontinuity, we noticed that SourceBuffer cannot append the buffers after the discontinuity tags successfully. This might be caused by audio streams getting trimmed if there's also video, based on the original timestamps.
https://github.com/videojs/mux.js/blob/master/lib/mp4/transmuxer.js#L1016

I've attached two pieces of contents clipped from a live stream playlist, one right before EXT-X-DISCONTINUITY tag, and the other one right after the tag for your reference.

[discontinuity.zip](https://github.com/videojs/mux.js/files/4421990/discontinuity.zip)

joeyparrish:
Because the first segment after a discontinuity may have a smaller timestamp than the first segment overall, the post-discontinuity segment may have its audio completely removed.

With the keepOriginalTimestamps flag, this subtle form of timestamp manipulation should be avoided, leaving apps in complete control of the timestamps.

Closes #328

### DIFF
--- a/test/transmuxer.test.js
+++ b/test/transmuxer.test.js
@@ -3486,8 +3486,9 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
 
   QUnit.test('Audio frames trimmed before video, ' + name, function() {
     var
-    segments = [],
+      segments = [],
       earliestDts = 15000,
+      baseTime = test.options.baseMediaDecodeTime || test.baseMediaSetter || 0,
       transmuxer = createTransmuxer();
 
     transmuxer.on('data', function(segment) {
@@ -3506,7 +3507,7 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
 
     transmuxer.push(packetize(audioPes([
       0x19, 0x47
-    ], true, earliestDts - (test.options.baseMediaDecodeTime || test.baseMediaSetter || 0) - 1)));
+    ], true, earliestDts - baseTime - 1)));
     transmuxer.push(packetize(videoPes([
       0x09, 0x01 // access_unit_delimiter_rbsp
     ], true, earliestDts)));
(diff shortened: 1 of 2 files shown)


gesinger:
@joeyparrish thanks for the PR! I'm looking into the test failure to see the best way to resolve it. Will keep you posted.

joeyparrish:
I tried updating the test so that the expectation depends on the flag, but that doesn't give me the results I expected.  I'm combination with the other options, it seems the flag has no effect on the output length in that test.  I'm not sure why.  Did I miss something in my change?

joeyparrish:
I finally figured out how to update the test correctly.  The other options have an effect on the output as well, effectively overriding keepOriginalTimestamps.  With the commit I just added, all tests are passing locally now, and I expect Travis to confirm soon.

Thanks!

test/transmuxer.test.js
```diff
@@ -3529,7 +3530,11 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
     QUnit.equal(segments.length, 1, 'generated a combined segment');
     // The audio frame is 10 bytes. The full data is 305 bytes without anything
     // trimmed. If the audio frame was trimmed this will be 295 bytes.
-    QUnit.equal(segments[0].data.length, 295, 'trimmed audio frame');
+    if (test.options.keepOriginalTimestamps && !baseTime) {
```
gesinger, line 3533:
```suggestion
// Note that if the baseMediaDecodeTime is set via options or the setter, frames may still
// be removed, even if keepOriginalTimestamps is true.
if (test.options.keepOriginalTimestamps && !baseTime) {
```

joeyparrish:
I appreciate the suggestion.  I did *not* know this from the code, only from debugging and examining the outputs in the test.  So I like it much better with the comment than without.  Thanks!

PR #330, opened 2020-04-06
fix: skip DTG1 packets in parseSei

JulianDomingo:
Atypical streams containing 2+ SEI NAL payloads of type 0x04 (CEA-708 caption content) in the sei_rbsp (specifically a DTG1 payload preceding a GA94 payload) fail to display captions in this scenario, as the current logic results in parseSei() prematurely returning with the DTG1 payload.

By checking for the user ID within parseSei, mux.js can continue iterating through the sei_rbsp for the GA94 payload.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -1010,10 +1010,11 @@ Transmuxer = function(options) {
           pipeline.videoSegmentStream = new VideoSegmentStream(videoTrack, options);
 
           pipeline.videoSegmentStream.on('timelineStartInfo', function(timelineStartInfo) {
-          // When video emits timelineStartInfo data after a flush, we forward that
-          // info to the AudioSegmentStream, if it exists, because video timeline
-          // data takes precedence.
-            if (audioTrack) {
+            // When video emits timelineStartInfo data after a flush, we forward that
+            // info to the AudioSegmentStream, if it exists, because video timeline
+            // data takes precedence.  Do not do this if keepOriginalTimestamps is set,
+            // because this is a particularly subtle form of timestamp alteration.
+            if (audioTrack && !options.keepOriginalTimestamps) {
               audioTrack.timelineStartInfo = timelineStartInfo;
               // On the first segment we trim AAC frames that exist before the
               // very earliest DTS we have seen in video because Chrome will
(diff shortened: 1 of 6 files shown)


joeyparrish:
Shaka Player is also interested in this PR.

joeyparrish:
@gesinger, could you look at this PR when you have time?  This is another fix we'd like to have in our next release of Shaka Player, if possible.

gkatsev:
Looking at the netlify build, I can see that 608 captions continue to work in bipbop but I don't quite understand what's happening.
A test would be helpful as well.

gkatsev:
looking at the spec and re-reading the description and code again, I guess what's happening is that we're only accounting for sei packets in the AFD format and exit early but it's possible that in ATSC streams, there's more data that we should parse through and then exit on the GA94 payload? At the same time, we don't want to accumulate the extra payloads because they won't be caption data. Am I understanding it correctly?
Feel free to point me at the appropriate specs and what not.

JulianDomingo:
I don't believe mux.js would accumulate extra payloads with this change (the returned payload wouldn't contain non-caption data). 
The intention is to have mux.js ensure it always returns the GA94 payload in the scenario where a stream contains both an sei packet of AFD and ATSC format. Currently, mux.js expects the stream to contain only one sei packet format. For example, mux.js would return the AFD payload if it appears first in the stream since it only checks for payload type 0x04 ("USER_DATA_REGISTERED_ITU_T_T35"). Visually:
... -> AFD_data (parseSei() stops here and returns) -> ATSC1_data -> ...
but should be:
... -> AFD_data (skip) -> ATSC1_data (parseSei() stops here and returns) -> ...

Spec-wise I don't see any issues, but this PR stemmed from an internal report where streams didn't display captions, and after running it through a media analyzer, they all contained both an AFD and ATSC sei packet. With the PR, captions were displayed properly.

gkatsev:
Thanks! It does seem like it's correct, I was asking about the specs mostly because I have very little knowledge on this subject matter.
Adding a test like https://github.com/videojs/mux.js/blob/master/test/caption-stream.test.js#L56-L91 would be super helpful, if you can.

JulianDomingo:
Sorry, please refer to https://github.com/videojs/mux.js/pull/333 instead.

PR #334, opened 2020-04-16
revert: use baseMediaDecode to set earliest dts (#317)

brandonocasey:
Reverts the part of #317 that always uses `baseMediaDecodeTime` to trim audio even when `keepOriginalTimestamps` is on. It seems like the actual fix fo the issue in #317 was fixed with #329. I have also included the updates from #329 into the partial transmuxer in this pull request so that it remains up to date. 

Right now this code breaks alternative audio tracks as we do not trim audio as expected and switching to an alternative audio track causes us to infinite buffer due to a gap.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -988,8 +988,6 @@ Transmuxer = function(options) {
     pipeline.elementaryStream.on('data', function(data) {
       var i;
 
-      var baseMediaDecodeTime = !options.keepOriginalTimestamps ? self.baseMediaDecodeTime : 0;
-
       if (data.type === 'metadata') {
         i = data.tracks.length;
 
@@ -997,10 +995,10 @@ Transmuxer = function(options) {
         while (i--) {
           if (!videoTrack && data.tracks[i].type === 'video') {
             videoTrack = data.tracks[i];
-            videoTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
+            videoTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
           } else if (!audioTrack && data.tracks[i].type === 'audio') {
             audioTrack = data.tracks[i];
-            audioTrack.timelineStartInfo.baseMediaDecodeTime = baseMediaDecodeTime;
+            audioTrack.timelineStartInfo.baseMediaDecodeTime = self.baseMediaDecodeTime;
           }
         }
 
@@ -1082,7 +1080,9 @@ Transmuxer = function(options) {
   this.setBaseMediaDecodeTime = function(baseMediaDecodeTime) {
     var pipeline = this.transmuxPipeline_;
 
-    this.baseMediaDecodeTime = baseMediaDecodeTime;
+    if (!options.keepOriginalTimestamps) {
+      this.baseMediaDecodeTime = baseMediaDecodeTime;
+    }
 
     if (audioTrack) {
       audioTrack.timelineStartInfo.dts = undefined;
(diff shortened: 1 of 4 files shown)


lib/partial/transmuxer.js
```diff
@@ -72,7 +72,7 @@ var tsPipeline = function(options) {
       pipeline.videoSegmentStream = new VideoSegmentStream(pipeline.tracks.video, options);
 
       pipeline.videoSegmentStream.on('timelineStartInfo', function(timelineStartInfo) {
-        if (pipeline.tracks.audio) {
+        if (pipeline.tracks.audio && !options.keepOriginalTimestamps) {
```
brandonocasey, line 75: The fix from #329 but for the partial transmuxer.

lib/mp4/transmuxer.js
```diff
@@ -1082,7 +1080,9 @@ Transmuxer = function(options) {
   this.setBaseMediaDecodeTime = function(baseMediaDecodeTime) {
     var pipeline = this.transmuxPipeline_;
 
-    this.baseMediaDecodeTime = baseMediaDecodeTime;
+    if (!options.keepOriginalTimestamps) {
```
gesinger, line 1083: Since this seems to change behavior, does it break https://github.com/videojs/mux.js/pull/329/files#diff-0f007c404c74edbc9d98f0a4d7f54f86R3533 ?

test/transmuxer.test.js
```diff
@@ -3530,10 +3530,8 @@ QUnit.test('adjusts caption and ID3 times when configured to adjust timestamps',
     QUnit.equal(segments.length, 1, 'generated a combined segment');
     // The audio frame is 10 bytes. The full data is 305 bytes without anything
     // trimmed. If the audio frame was trimmed this will be 295 bytes.
-// Note that if the baseMediaDecodeTime is set via options or the setter, frames may still
-// be removed, even if keepOriginalTimestamps is true.
-if (test.options.keepOriginalTimestamps && !baseTime) {
-      QUnit.equal(segments[0].data.length, 305, 'trimmed audio frame');
+    if (test.options.keepOriginalTimestamps && !baseTime) {
```
brandonocasey, line 3533: spacing issues

test/partial.test.js
```diff
@@ -128,9 +129,14 @@ QUnit.module('Partial Transmuxer - Options');
     transmuxer.flush();
 
     // the partial transmuxer only generates a video segment
-    // when all audio frames are trimmed. So we should only have
-    // a video segment
-    QUnit.equal(segments.length, 1, 'generated only a video segment');
-    QUnit.equal(segments[0].type, 'video', 'segment is video');
+    // when all audio frames are trimmed.
```
brandonocasey, line 132: similar change as the main tests for the partial tests.

lib/mp4/transmuxer.js
brandonocasey: I removed that comment. This does changes behavior but only very recent behavior, that I believe wasn't the real fix for what we needed.

lib/partial/transmuxer.js
```diff
@@ -191,7 +192,7 @@ var aacPipeline = function(options) {
 
     pipeline.tracks.audio = pipeline.tracks.audio || {
       timelineStartInfo: {
-        baseMediaDecodeTime: !options.keepOriginalTimestamps ? options.baseMediaDecodeTime : 0
+        baseMediaDecodeTime: options.baseMediaDecodeTime
```
gesinger, line 195: Do we still want to allow setting of `baseMediaDecodeTime` if `keepOriginalTimestamps` is true?

brandonocasey: We do in the main transmuxer, so we do in the partial as well.

PR #336, opened 2020-04-27
fix: 708 caption error caused by non GA94 caption

brandonocasey:
See the code for when this happens.

### DIFF
--- a/lib/tools/caption-packet-parser.js
+++ b/lib/tools/caption-packet-parser.js
@@ -62,17 +62,16 @@ var parseSei = function(bytes) {
     // this sei_message is a 608/708 caption so save it and break
     // there can only ever be one caption message in a frame's sei
     if (!result.payload && payloadType === USER_DATA_REGISTERED_ITU_T_T35) {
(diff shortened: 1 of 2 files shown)


lib/tools/caption-packet-parser.js
```diff
@@ -62,17 +62,16 @@ var parseSei = function(bytes) {
     // this sei_message is a 608/708 caption so save it and break
     // there can only ever be one caption message in a frame's sei
     if (!result.payload && payloadType === USER_DATA_REGISTERED_ITU_T_T35) {
-      result.payloadType = payloadType;
```
brandonocasey, line 65: previously `payloadType` and `payloadSize` would stay set even though we were nulling out `payload` this caused https://github.com/videojs/mux.js/blob/master/lib/tools/caption-packet-parser.js#L95 via https://github.com/videojs/mux.js/blob/master/lib/m2ts/caption-stream.js#L60-L65 to throw an error as `payload` was undefined but `payloadType` was `USER_DATA_REGISTERED_ITU_T_T35`

PR #345, opened 2020-06-11
refactor: dedupe code into modules for smaller vhs

brandonocasey:
See https://github.com/videojs/http-streaming/pull/863

### DIFF
--- a/lib/mp4/probe.js
+++ b/lib/mp4/probe.js
@@ -164,18 +164,28 @@ compositionStartTime = function(timescales, fragment) {
     // The spec states that track run samples contained within a `traf` box are contiguous, but
     // it does not explicitly state whether the `traf` boxes themselves are contiguous.
     // We will assume that they are, so we only need the first to calculate start time.
-    var tfhd = parseTfhd(findBox(trafBoxes[0], ['tfhd'])[0]);
-    var trun = parseTrun(findBox(trafBoxes[0], ['trun'])[0]);
-    var tfdt = parseTfdt(findBox(trafBoxes[0], ['tfdt'])[0]);
+    var tfhd = findBox(trafBoxes[0], ['tfhd'])[0];
+    var trun = findBox(trafBoxes[0], ['trun'])[0];
+    var tfdt = findBox(trafBoxes[0], ['tfdt'])[0];
 
     if (tfhd) {
-      trackId = tfhd.trackId;
+      var parsedTfhd = parseTfhd(tfhd);
+
+      trackId = parsedTfhd.trackId;
     }
+
     if (tfdt) {
-      baseMediaDecodeTime = tfdt.baseMediaDecodeTime;
+      var parsedTfdt = parseTfdt(tfdt);
+
+      baseMediaDecodeTime = parsedTfdt.baseMediaDecodeTime;
     }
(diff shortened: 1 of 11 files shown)


lib/mp4/probe.js
```diff
@@ -160,16 +164,18 @@ compositionStartTime = function(timescales, fragment) {
...
-      if (parsedTraf.boxes[i].type === 'tfhd') {
-        trackId = parsedTraf.boxes[i].trackId;
-      } else if (parsedTraf.boxes[i].type === 'tfdt') {
-        baseMediaDecodeTime = parsedTraf.boxes[i].baseMediaDecodeTime;
-      } else if (parsedTraf.boxes[i].type === 'trun' && parsedTraf.boxes[i].samples.length) {
-        compositionTimeOffset = parsedTraf.boxes[i].samples[0].compositionTimeOffset || 0;
-      }
+    var tfhd = parseTfhd(findBox(trafBoxes[0], ['tfhd'])[0]);
+    var trun = parseTrun(findBox(trafBoxes[0], ['trun'])[0]);
+    var tfdt = parseTfdt(findBox(trafBoxes[0], ['tfdt'])[0]);
```
gesinger, line 169: Not sure if this is a case we will/have run into, but it looks like if those boxes are not in the segment, we will throw exceptions here (findBox will return an empty array, first element will be `undefined` and each parse function would try to reference the first element of `undefined`). In the past it seemed we just didn't have the result.

lib/mp4/probe.js
```diff
@@ -160,15 +164,27 @@ compositionStartTime = function(timescales, fragment) {
...
-      } else if (parsedTraf.boxes[i].type === 'trun' && parsedTraf.boxes[i].samples.length) {
-        compositionTimeOffset = parsedTraf.boxes[i].samples[0].compositionTimeOffset || 0;
+    var tfhd = findBox(trafBoxes[0], ['tfhd'])[0];
+    var trun = findBox(trafBoxes[0], ['trun'])[0];
+    var tfdt = findBox(trafBoxes[0], ['tfdt'])[0];
+
+    if (tfhd) {
+      var parsedTfhd = parseTfhd(tfhd);
+
+      trackId = tfhd.trackId;
```
gesinger, line 174: I think this should be parsedTfhd

lib/mp4/probe.js
```diff
@@ -160,15 +164,27 @@ compositionStartTime = function(timescales, fragment) {
...
+    }
+
+    if (tfdt) {
+      var parsedTfdt = parseTfdt(tfdt);
+
+      baseMediaDecodeTime = parseTfdt.baseMediaDecodeTime;
```
gesinger, line 180: I think this should be parsedTfdt

lib/mp4/probe.js
```diff
@@ -160,15 +164,27 @@ compositionStartTime = function(timescales, fragment) {
...
+    }
+
+    if (trun) {
+      var parsedTrun = parseTrun(trun);
+
+      if (parseTrun.samples && parseTrun.samples.length) {
+        compositionTimeOffset = parseTrun.samples[0].compositionTimeOffset || 0;
```
gesinger, line 187: I think these should be parsedTrun

PR #346, opened 2020-07-08
feat: Cea708 caption decoder

iamstickfigure:
This adds another caption decoder for 708 captions along side the existing 608 caption decoder. It isn't fully featured, but it has almost as many features as the existing 608 decoder. 

Like the 608 decoder, this does not support positioning.
Unlike the 608 decoder, this does not support italics and underline.


gkatsev:
Thank you so much! We're swamped trying to get [VHS 2.1](https://github.com/videojs/http-streaming/milestone/4) out, but we'll review as soon as we can!

gkatsev:
Also, would be awesome to get some tests added.

iamstickfigure:
Regarding tests, I actually have a test harness set up locally that I've been using while iterating on my code. However, it relies on the existence of reading in large TS files. I took a look at the 608 decoder tests, and those tests do not read in source video files, so I'm not entirely sure how I would adapt my test harness in a way that avoids having large TS files in the repo.

Looking again though, I see a file already present called `mixed-608-708-captions.js`. I suppose I could use that assuming it actually has 708 captions in the data. I'll take a closer look at that.

iamstickfigure:
> Also, would be awesome to get some tests added.

I have added some tests

gkatsev:
Hey, just wanted to post an update since it's been a while. We've been swamped but this is on our plan to look into really soon.

iamstickfigure:
Awesome. No problem. It gave us some extra time to discover and fix a few bugs, add some improvements, and implement a bit more of the spec like roll-up captions.

gkatsev:
Speaking of roll-up captions, I just finished implementing region support in vtt.js https://github.com/videojs/vtt.js/pull/50, which was added to webvtt specifically for 608/708 rollup support.

iamstickfigure:
Awesome. For now, my roll-up implementation just involves replacing the first row with the second row, and repeat. So it's like a jumpy version of roll-up captions, but it did the trick. It might be worth converting over to using vtt regions at some point in a separate PR. I assume this enables some sort of animated scrolling between rows?

gkatsev:
yup, exactly. It transitions each cue up when they are in the same region.

brandonocasey (approved):

gesinger (approved):

gkatsev:
Thanks @iamstickfigure. We'll get it into a release soon!

PR #347, opened 2020-07-09
fix: ID3 does not indicate aac data

brandonocasey:
ID3 tags are not specific to aac data. Almost any audio container format can use it. AAC has a specific syncbyte `0xFFF1` that we need to look for instead.

### DIFF
--- a/lib/aac/utils.js
+++ b/lib/aac/utils.js
@@ -24,13 +24,35 @@ var ADTS_SAMPLING_FREQUENCIES = [
   7350
 ];
 
+var parseId3TagSize = function(header, byteIndex) {
+  var
+    returnSize = (header[byteIndex + 6] << 21) |
+                 (header[byteIndex + 7] << 14) |
+                 (header[byteIndex + 8] << 7) |
+                 (header[byteIndex + 9]),
+    flags = header[byteIndex + 5],
+    footerPresent = (flags & 16) >> 4;
+
+  if (footerPresent) {
+    return returnSize + 20;
+  }
+  return returnSize + 10;
+};
+
+// TODO: use vhs-utils
 var isLikelyAacData = function(data) {
+  var offset = 0;
+
   if ((data[0] === 'I'.charCodeAt(0)) &&
       (data[1] === 'D'.charCodeAt(0)) &&
       (data[2] === '3'.charCodeAt(0))) {
-    return true;
+    offset = parseId3TagSize(data, 0);
   }
-  return false;
+
+  return data.length >= offset + 2 &&
+    (data[offset] & 0xFF) === 0xFF &&
+    (data[offset + 1] & 0xE0) === 0xE0 &&
+    (data[offset + 1] & 0x16) === 0x10;
 };
 
 var parseSyncSafeInteger = function(data) {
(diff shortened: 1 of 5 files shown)


lib/aac/utils.js
```diff
@@ -24,13 +24,35 @@ var ADTS_SAMPLING_FREQUENCIES = [
   7350
 ];
 
+var parseId3TagSize = function(header, byteIndex) {
```
brandonocasey, line 27: Code is unchanged. I just moved it above `isLikelyAacData` as we use it there now.

lib/aac/utils.js
```diff
@@ -24,13 +24,35 @@ var ADTS_SAMPLING_FREQUENCIES = [
...
+
+// TODO: use vhs-utils
 var isLikelyAacData = function(data) {
+  var offset = 0;
+
   if ((data[0] === 'I'.charCodeAt(0)) &&
       (data[1] === 'D'.charCodeAt(0)) &&
       (data[2] === '3'.charCodeAt(0))) {
-    return true;
+    offset = parseId3TagSize(data, 0);
```
gesinger, line 49: Should we check the length here before this call?

test/aac-utils.test.js
```diff
@@ -15,8 +15,9 @@ QUnit.test('correctly determines aac data', function(assert) {
   assert.ok(utils.isLikelyAacData(testSegment), 'test segment is aac');
 
   var id3Offset = utils.parseId3TagSize(testSegment, 0);
-
   assert.ok(utils.isLikelyAacData(testSegment.subarray(id3Offset)), 'test segment is aac without id3');
+  assert.notOk(utils.isLikelyAacData(testSegment.subarray(id3Offset + 25)), 'non aac data not recognized');
+  assert.notOk(utils.isLikelyAacData(testSegment.subarray(0, 5)), 'not enouch aac data is not recognized');
```
gesinger, line 20:
```suggestion
  assert.notOk(utils.isLikelyAacData(testSegment.subarray(0, 5)), 'not enough aac data is not recognized');
```

lib/aac/utils.js
```diff
@@ -24,13 +24,35 @@ var ADTS_SAMPLING_FREQUENCIES = [
...
   }
-  return false;
+
+  return data.length >= offset + 2 &&
+    (data[offset] & 0xFF) === 0xFF &&
+    (data[offset + 1] & 0xE0) === 0xE0 &&
```
gesinger, line 54: Since the ADTS header syncword is 0xFFF, should this be 0xF0 instead of 0xE0?

lib/aac/utils.js
```diff
@@ -24,13 +24,35 @@ var ADTS_SAMPLING_FREQUENCIES = [
...
+    (data[offset + 1] & 0x16) === 0x10;
```
gesinger, line 55: May be worth a comment above that mentions the syncword that's being checked for and that this comparison verifies the 2 layer bits as always 0.

lib/aac/utils.js
gesinger: May be worth a comment.

lib/aac/utils.js
```diff
@@ -51,6 +51,8 @@ var isLikelyAacData = function(data) {
 
   return data.length >= offset + 2 &&
     (data[offset] & 0xFF) === 0xFF &&
+    // verify that the 2 layer bits are 0, aka this
```
gesinger, line 54: Should this comment be on the next line?

brandonocasey: done

PR #350, opened 2020-10-05
fix: handle recursive id3 in isLikelyAacData

brandonocasey:
handle multiple id3 sections at the start of aac data.

### DIFF
--- a/lib/aac/utils.js
+++ b/lib/aac/utils.js
@@ -45,7 +45,7 @@ var parseId3TagSize = function(header, byteIndex) {
 var getId3Offset = function(data, offset) {
   if (data.length - offset < 10 ||
       data[offset] !== 'I'.charCodeAt(0) ||
-      data[offset + 1] !== 'D'.charCodeAt(0) &&
+      data[offset + 1] !== 'D'.charCodeAt(0) ||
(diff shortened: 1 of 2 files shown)


lib/aac/utils.js
```diff
@@ -39,15 +39,23 @@ var parseId3TagSize = function(header, byteIndex) {
...
 
+var getId3Offset = function(data, offset) {
+  if (data.length - offset < 10 ||
+      data[offset] !== 'I'.charCodeAt(0) ||
+      data[offset + 1] !== 'D'.charCodeAt(0) &&
+      data[offset + 2] !== '3'.charCodeAt(0)) {
+    return offset;
+  }
+
+  offset += parseId3TagSize(data, offset);
```
gesinger, line 50: I think it might be too much of an edge case to worry about, but technically, if the data we pass in starts with ID3 but isn't ID3 data, and the indexes for the returnSize in the header equals -10, then we will get stuck in an infinite loop.

brandonocasey: I don't think the return size can be negative, since we only use Uint8Array's for our data. If we used Int8Array then it might be a problem, not sure it's worth worrying about.

gesinger (approved):

PR #365, opened 2021-01-15
feat(CaptionStream): add flag to turn off 708 captions

gkatsev:
This options flow through the transmuxer into CaptionStream.

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -29,6 +29,7 @@ var CaptionStream = function(options) {
   this.parse708captions_ = typeof options.parse708captions === 'boolean' ?
     options.parse708captions :
     true;
+  this.parse708captions_ = true;
 
   this.captionPackets_ = [];
 
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -24,8 +24,10 @@ var CaptionStream = function(options) {
 
   CaptionStream.prototype.init.call(this);
 
-  // cc708 flag, default to true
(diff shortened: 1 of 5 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -20,10 +20,13 @@
 var Stream = require('../utils/stream');
 var cea708Parser = require('../tools/caption-packet-parser');
 
-var CaptionStream = function() {
+var CaptionStream = function(options) {
 
   CaptionStream.prototype.init.call(this);
 
+  // cc708 flag, default to true
+  this.cc708_ = typeof options.cc708 === 'boolean' ? options.cc708 : true;
```
gesinger, line 28: I think it might be better to add a bit more description. Maybe something like `include708Captions` or `parse708Captions`.

gesinger, line 28: Might be worth changing the name to be a bit more descriptive. Maybe `include708Captions` or `parse708Captions`.

gesinger (commented):
Probably worth adding a test for the option.

lib/m2ts/caption-stream.js
```diff
@@ -20,10 +20,17 @@
...
+var CaptionStream = function(options) {
+  options = options || {};
 
   CaptionStream.prototype.init.call(this);
 
+  // parse708captions flag, default to true
+  this.parse708captions_ = typeof options.parse708captions === 'boolean' ?
+    options.parse708captions :
+    true;
+  this.parse708captions_ = true;
```
gkatsev, line 32: this is to demonstrate the test failing

gesinger (approved):

PR #366, opened 2021-01-19
feat: add a nodejs binary for transmux via command line

brandonocasey:


bin/transmux.js
```diff
@@ -0,0 +1,125 @@
...
+  for (let i = 0; i < args.length; i++) {
+    const arg = args[i];
+
+    if ((/^--version|-v$/).test(arg)) {
+      console.log(`transmux.js v${version}`);
+      process.exit(0);
+    } else if ((/^--help|-h$/).test(arg)) {
+      showHelp();
+      process.exit(0);
+    } else if ((/^--debugger|-d$/).test(arg)) {
```
gkatsev, line 38: we currently error out if `muxjs-transmux` is run without arguments, we probably should default to `showHelp()`

gkatsev (commented):
Seems to create mp4 files but the timing is all messed up. Is it because there's no associated init segment?

brandonocasey:
The timing issues could be because I set base media decode time to a default value of 100000

brandonocasey:
Actually I think you mean the duration of the file when playing back? That happens because we set duration to `0xffffffff` by default in our generated fmp4 files.

gkatsev:
Yeah, of the generated mp4 file.

gkatsev (approved):
I think I'm still getting weird durations in the output but I think we can fix it later.

PR #375, opened 2021-02-18
parse edts box implemented

nklhrstv:
Hi,
Ive implemented parsing for edts atom by following this [spec](https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html)


Ivshti (approved):

nklhrstv:
Hi @gkatsev @gesinger @joeyparrish
Any chances for a review any time soon? I need this in my project.

test/utils/mp4-helpers.js
```diff
@@ -109,6 +109,14 @@ module.exports.sampleMoov =
...
               0x01, 0x2c, 0x00, 0x00, // 300 in 16.16 fixed-point
               0x00, 0x96, 0x00, 0x00), // 150 in 16.16 fixed-point
+          box('edts',
+              box('elst',
+                0x00, // version
+                0x00, 0x00, 0x00, // flags
+                0x00, 0x00, 0x00, 0x01, // entry_count
+                0x00, 0x00, 0x00, 0x00, // segment_duration
+                0x00, 0x00, 0x04, 0x00, // media_time
+                0x00, 0x01, 0x00, 0x00)), // media_rate
```
gesinger, line 119: Since it requires a bit more math for the fixed point, it may be worth having a fractional value.

gesinger (commented):
Thank you for the PR @nklhrstv !

Looks good judging by the layout I see in https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html#//apple_ref/doc/uid/TP40000939-CH204-32975

One minor comment.

test/utils/mp4-helpers.js
nklhrstv: good catch. ive changed test value for media rate

gesinger (approved):
Thank you!

lib/tools/mp4-inspector.js
```diff
@@ -123,6 +123,30 @@ var
...
+        result = {
+          version: view.getUint8(0),
+          flags: new Uint8Array(data.subarray(1, 4)),
+          edits: []
+        },
+        entryCount = view.getUint32(4),
+        i;
+      for (i = 8; entryCount; i += 12, entryCount--) {
+        result.edits.push({
+          segmentDuration: view.getUint32(i),
```
brandonocasey, line 143: looking at the mp4 spec it seems like we need to look at `result.version` and if `result.version` is `0` then `segmentDuration` and `mediaTime` will 32 bit and if `result.version` is `1` then segmentDuration` and `mediaTime`  will be 64 bit.

brandonocasey: for an example of grabbing 64 bit values (i know js doesn't actually support that many bits for numbers, but we can still get them up to 52 bits) see https://github.com/videojs/mux.js/blob/main/lib/tools/parse-sidx.js#L20-L21

nklhrstv: What about BigInt? Ive used `.getBigUint64`.

nklhrstv: What is the minimum supported node version for this library?

gkatsev: IE11 😂 (sorry)

nklhrstv: Check it now

nklhrstv: Ive added 64 bit test case. 
should be all good now

brandonocasey (approved):

PR #379, opened 2021-03-10
parse ctts atom implemented

nklhrstv:
Implementation of ctts atom parsing.
@gkatsev @gesinger @joeyparrish could you review this PR? I need this atom in my project.


joeyparrish:
I'm not one of the maintainers of this project, so I won't be able to approve the PR.  Sorry for any confusion!

joeyparrish:
I'm not deeply familiar with norms and preferred style in mux.js, and I haven't compared your parser to any spec for the ctts box, but the implementation looks reasonable to me, for what that's worth.

gkatsev (approved):
Style-wise, it matches existing code. Also, appreciate that there are tests! Appreciate you looking at this @joeyparrish.
I'll have to defer to @gesinger or @brandonocasey to do a deeper review.

joeyparrish:
@gkatsev, if you are happy with the project-specific details, and it's just a question of correctness, I found docs from Apple detailing the ctts box, and the parser matches perfectly as far as I can tell:

https://developer.apple.com/library/archive/documentation/QuickTime/QTFF/QTFFChap2/qtff2.html#//apple_ref/doc/uid/TP40000939-CH204-SW40

joeyparrish (approved):
Approved by an interested outsider, for whatever that's worth.  :-D

lib/tools/mp4-inspector.js
```diff
@@ -425,6 +425,24 @@ var
...
+          version: view.getUint8(0),
+          flags: new Uint8Array(data.subarray(1, 4)),
+          compositionOffsets: []
+        },
+        entryCount = view.getUint32(4),
+        i;
+      for (i = 8; entryCount; i += 8, entryCount--) {
+        result.compositionOffsets.push({
+          sampleCount: view.getUint32(i),
+          sampleOffset: view.getInt32(i + 4)
```
brandonocasey, line 441: in the spec for mp4 `version 0` of this box uses an `uint` while `version 1` uses an int. Would you mind making the change for this case?

nklhrstv:
@joeyparrish Sorry, i thought you are a maintainer

lib/tools/mp4-inspector.js
nklhrstv: can you please give me a reference for ctts spec?
[bento4](https://github.com/axiomatic-systems/Bento4/blob/be15647cc9750489cb2972f630d099b6426f03a7/Source/C%2B%2B/Core/Ap4CttsAtom.cpp#L102) actually uses only uint and used to have checks for quick time support
as far as i understand negative offsets ware used only in quicktime and they ware never part of the spec

lib/tools/mp4-inspector.js
```diff
@@ -425,6 +425,24 @@ var
...
+          version: view.getUint8(0),
+          flags: new Uint8Array(data.subarray(1, 4)),
+          compositionOffsets: []
+        },
+        entryCount = view.getUint32(4),
+        i;
+      for (i = 8; entryCount; i += 8, entryCount--) {
+        result.compositionOffsets.push({
+          sampleCount: view.getUint32(i),
+          sampleOffset: view.getUint32(i + 4)
```
brandonocasey, line 441: The spec is closed/paid for so I can't really link a reference to it. Basically you have to check the box version for this box and use `view.getUint32` if it is version 0 and `view.getInt32` for version 1. Should be something like this suggestion.
```suggestion
          sampleOffset: view[(result.version === 0 ? 'getUint32' :'getInt32')](i + 4)
```

gkatsev:
We definitely value @joeyparrish's opinion!

joeyparrish:
> We definitely value @joeyparrish's opinion!

Then I've fooled you all!!  😈

lib/tools/mp4-inspector.js
nklhrstv: Done

brandonocasey (approved):

PR #380, opened 2021-03-16
stss atom parsing

nklhrstv:
parse stss atom implemented


brandonocasey (approved):
verified this against the spec and it looks like what I would expect.

nklhrstv:
When should i exprect to get it merged?

PR #383, opened 2021-05-17
feat(BREAKING): use bigint for 64 bit ints if needed and available.

brandonocasey:
Stop reading numbers imprecisely by using bigint (if needed).

EDIT: the only way I can see this being non-breaking is if we add an option to the function to use bigint if available, but i'm not sure if that is worth it.


lib/mp4/mp4-generator.js
```diff
@@ -581,8 +582,8 @@ traf = function(track) {
     0x00, 0x00, 0x00, 0x00  // default_sample_flags
   ]));
 
-  upperWordBaseMediaDecodeTime = Math.floor(track.baseMediaDecodeTime / (UINT32_MAX + 1));
-  lowerWordBaseMediaDecodeTime = Math.floor(track.baseMediaDecodeTime % (UINT32_MAX + 1));
+  upperWordBaseMediaDecodeTime = Math.floor(track.baseMediaDecodeTime / (MAX_UINT32));
```
brandonocasey, line 585: We subtracted 1 from it at the top, only to minus 1 later...

lib/mp4/probe.js
```diff
@@ -87,52 +88,55 @@ timescale = function(init) {
  * fragment, in seconds
  */
 startTime = function(timescale, fragment) {
-  var trafs, baseTimes, result;
+  var trafs, result;
```
brandonocasey, line 91: This code was all kinds of weird. We create an empty array twice to map over another array of trafs?? Then we map over all tfhd/tfdt and parse them, but only take the first? It's all kinds of nuts. 

I switched it so that we loop over all `traf`s, and only grab the lowest time that we find (which is what we ultimately did later in the code). Furthermore we only look for one `tfhd` and one `tfdt` as there will only ever be one, and we ignored other in the old code anyway.

lib/mp4/probe.js
```diff
@@ -87,52 +88,55 @@ timescale = function(init) {
...
-  }));
+    // convert base time to seconds if it is a valid number.
+    let seconds;
+    if (typeof baseTime === 'bigint') {
+      seconds = baseTime / globalThis.BigInt(scale);
+    } else if (typeof baseTime === 'number' && !isNaN(baseTime)) {
+      seconds = baseTime / scale;
+    }
+
+    if (seconds < Number.MAX_SAFE_INTEGER) {
```
brandonocasey, line 128: if converting it to the appropriate scale, brings it below the threshold to be a bigInt bring it back to a number.

test/mp4-inspector.test.js
```diff
@@ -403,7 +404,7 @@ QUnit.test('can parse a moov', function(assert) {
           flags: new Uint8Array([0, 0, 0]),
           edits: [{
             segmentDuration: 0,
-            mediaTime: 1152921504606847000,
+            mediaTime: BigInt('0x1000000000000000'),
```
brandonocasey, line 407: This number was 100% wrong before and relying on consistent incorrect rounding in javascript.

gkatsev (approved):

PR #386, opened 2021-06-17
fix: inspect all program map tables for stream types

brandonocasey:
We currently have issues with ts files that add stream types after the first program map table as we only look at the first one. So if audio were to be added in a subsequent program map table we wouldn't know to account for it.

### DIFF
--- a/lib/tools/ts-inspector.js
+++ b/lib/tools/ts-inspector.js
@@ -38,22 +38,13 @@ var parsePsi_ = function(bytes, pmt) {
 
       switch (type) {
         case 'pat':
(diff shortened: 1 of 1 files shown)


gkatsev (approved):
Confirmed working.

gkatsev:
Is vhs-utils needed in addition to this for VHS?

brandonocasey:
we do not need the vhs-utils update as we don't actually use the code there yet. I just try to keep the fixes between the two in sync as sometime soon we will want to switch mux.js to using vhs-utils.

PR #388, opened 2021-06-23
fix: on flush if a pmt has not been emitted and we have one, emit it

brandonocasey:
Always emit trackinfo, even if we didn't process a pmt.

### DIFF
--- a/lib/m2ts/m2ts.js
+++ b/lib/m2ts/m2ts.js
@@ -294,6 +294,7 @@ TransportParseStream.STREAM_TYPES  = {
 ElementaryStream = function() {
   var
     self = this,
+    segmentHadPmt = false,
     // PES packet fragments
     video = {
       data: [],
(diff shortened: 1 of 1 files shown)


lib/m2ts/m2ts.js
```diff
@@ -406,6 +407,40 @@ ElementaryStream = function() {
...
       }
 
+      if (!segmentHadPmt) {
+        var
+          pmt = {
+            type: 'metadata',
+            tracks: []
+          };
+        // translate audio and video streams to tracks
+        if (programMapTable.video !== null) {
```
gkatsev, line 417: looks like we need to null check `programMapTable` itself here and below too, likely.

brandonocasey: good call, I also realized that flushStreams isn't only called on `flush` which gave us a lot more `trackinfo` events than we wanted. I moved this code to `this.flush` and the unit tests in vhs pass 💯

gkatsev (approved):
Verified

PR #390, opened 2021-06-29
fix: prevent skipping frames on garbage adts data

brandonocasey:
Right now we run into an issue if adts contains data between sync words that is not part of a frame. 

Example
We have an adts file where bytes 0-5 are garbage, and after that the file is like normal. This is what would happen right now:
1. bytes 0-5 are garbage so they fail the syncword check and `i` is now incremented up to `5`.
2. We parse that frame successfully as 400 bytes long.
3. Then we remove those 400 bytes from the front of the buffer using subarray, but we do not reset `i` to `0`
4. On the next loop we check `buffer[i]` for a syncword but `i` is still `5` so we skip everything else in the buffer until the next syncword and further aggravate the problem by incrementing `i` some more.

### DIFF
--- a/lib/codecs/adts.js
+++ b/lib/codecs/adts.js
@@ -76,8 +76,8 @@ AdtsStream = function(handlePartialSegments) {
     // unpack any ADTS frames which have been fully received
     // for details on the ADTS header, see http://wiki.multimedia.cx/index.php?title=ADTS
 
-    // we use i + 7 here because we want to be able to parse the entire header.
-    // If we don't have enough bytes to do that, than we definitely won't have a full frame.
+    // We use i + 7 here because we want to be able to parse the entire header.
+    // If we don't have enough bytes to do that, then we definitely won't have a full frame.
     while ((i + 7) < buffer.length) {
       // Look for the start of an ADTS header..
       if ((buffer[i] !== 0xFF) || (buffer[i + 1] & 0xF6) !== 0xF0) {
--- a/lib/codecs/adts.js
+++ b/lib/codecs/adts.js
@@ -79,13 +79,23 @@ AdtsStream = function(handlePartialSegments) {
     // we use i + 5 here because we want to be able to parse the frame length
     // from the header. If we don't have enough bytes to do that, than we
     // definitely won't have a full frame.
+    let skipping;
     while ((i + 5) < buffer.length) {
       // Look for the start of an ADTS header..
       if ((buffer[i] !== 0xFF) || (buffer[i + 1] & 0xF6) !== 0xF0) {
         // If a valid header was not found,  jump one forward and attempt to
         // find a valid ADTS header starting at the next byte
+        if (!skipping) {
+          skipping = i;
+        }
         i++;
         continue;
+      }
+
+      if (typeof skipping === 'number') {
+        console.log(skipping, i);
+        console.log('skipped', buffer.subarray(skipping, i - 1));
+        skipping = null;
       }
 
       // The protection skip bit tells us if we have 2 bytes of CRC data at the
@@ -94,6 +104,7 @@ AdtsStream = function(handlePartialSegments) {
 
       // Frame length is a 13 bit integer starting 16 bits from the
       // end of the sync sequence
+      // NOTE: frame length includes the size of the header
       frameLength = ((buffer[i + 3] & 0x03) << 11) |
         (buffer[i + 4] << 3) |
         ((buffer[i + 5] & 0xe0) >> 5);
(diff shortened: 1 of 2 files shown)


lib/codecs/adts.js
```diff
@@ -119,20 +120,15 @@ AdtsStream = function(handlePartialSegments) {
...
-        return;
-      }
-
-      // Remove the finished frame from the buffer and start the process again
-      buffer = buffer.subarray(frameEnd);
+      i += frameLength;
     }
+
+    // remove processed bytes from the buffer.
+    buffer = buffer.subarray(i);
```
brandonocasey, line 131: remove the bytes that we processed from the buffer.

lib/codecs/adts.js
```diff
@@ -75,8 +75,11 @@ AdtsStream = function(handlePartialSegments) {
 
     // unpack any ADTS frames which have been fully received
     // for details on the ADTS header, see http://wiki.multimedia.cx/index.php?title=ADTS
-    while (i + 5 < buffer.length) {
 
+    // we use i + 3 here because we want to be able to parse the frame length
+    // from the header. If we don't have enough bytes to do that, than we
+    // definitely won't have a full frame.
+    while ((i + 3) < buffer.length) {
```
gesinger, line 82: Since we use i + 5 in determining frameLength below (and i + 6 below that), wouldn't we require at least that much?

lib/codecs/adts.js
```diff
@@ -64,7 +64,7 @@ AdtsStream = function(handlePartialSegments) {
 
     // Prepend any data in the buffer to the input data so that we can parse
     // aac frames the cross a PES packet boundary
-    if (buffer) {
+    if (buffer && buffer.length) {
```
brandonocasey, line 67: Buffer is no longer set to undefined if we don't have bytes left, I was torn on initializing it to an empty uint8 array and always concating with buffer to make all of this code simpler.

lib/codecs/adts.js
```diff
@@ -119,20 +120,15 @@ AdtsStream = function(handlePartialSegments) {
         samplingfrequencyindex: (buffer[i + 2] & 0x3c) >>> 2,
         // assume ISO/IEC 14496-12 AudioSampleEntry default of 16
         samplesize: 16,
-        data: buffer.subarray(i + 7 + protectionSkipBytes, frameEnd)
+        data: buffer.subarray(i + 7 + protectionSkipBytes, i + frameLength)
```
gesinger, line 123: Would it make sense to keep the `frameEnd` variable and use it here and set `i` to equal `frameEnd` below (and above) just for clarity?

lib/codecs/adts.js
brandonocasey: you're right, I only looked at the first byte that it asks for, I will update this.

lib/codecs/adts.js
brandonocasey: From my point of view frameEnd just abstracts away what we are doing needlessly. I think that we should work in terms relative to i, rather than using an absolute value.

gesinger: 👍

lib/codecs/adts.js
```diff
@@ -99,12 +103,10 @@ AdtsStream = function(handlePartialSegments) {
...
 
-      frameEnd = i + frameLength;
-
-      // If we don't have enough data to actually finish this ADTS frame, return
-      // and wait for more data
-      if (buffer.byteLength < frameEnd) {
-        return;
+      // If we don't have enough data to actually finish this ADTS frame,
+      // then we have to wait for more data
+      if ((buffer.byteLength - i) < frameLength) {
```
gesinger, line 108: We should probably move this check up to just after determining the frame length (before the sample count) to prevent errors addressing `i + 6`

lib/codecs/adts.js
gesinger: I think the `frameEnd` variable can be removed in that case.

lib/codecs/adts.js
brandonocasey: changed the while loop condition to `i + 7` so that we won't even run the loop if we don't have enough bytes. That way and we can keep all the parsing logic together.

lib/codecs/adts.js
```diff
@@ -75,8 +75,10 @@ AdtsStream = function(handlePartialSegments) {
 
     // unpack any ADTS frames which have been fully received
     // for details on the ADTS header, see http://wiki.multimedia.cx/index.php?title=ADTS
-    while (i + 5 < buffer.length) {
 
+    // we use i + 7 here because we want to be able to parse the entire header.
+    // If we don't have enough bytes to do that, than we definitely won't have a full frame.
```
gesinger, line 80:
```suggestion
    // We use i + 7 here because we want to be able to parse the entire header.
    // If we don't have enough bytes to do that, then we definitely won't have a full frame.
```

gesinger (approved):

PR #392, opened 2021-07-08
refactor: rename warn event to log, change console logs to log events

brandonocasey:


lib/mp4/caption-parser.js
```diff
@@ -452,4 +460,6 @@ var CaptionParser = function() {
   this.reset();
 };
 
+CaptionParser.prototype = new Stream();
```
brandonocasey, line 463: Had to make this a stream as we don't have a better way to get listeners right now.

test/caption-parser.test.js
```diff
@@ -10,6 +10,8 @@ var dashInit = segments['dash-608-captions-init.mp4']();
 // This file includes 2 segments data to force a flush
 // of the first caption. The second caption is at 200s
 var dashSegment = segments['dash-608-captions-seg.m4s']();
+var malformedSei = segments['malformed-sei.m4s']();
+var malformedSeiInit = segments['malformed-sei-init.mp4']();
```
brandonocasey, line 14: Added these from the old issue, we should have done it then 🤷🏻

lib/mp4/caption-parser.js
```diff
@@ -97,8 +98,7 @@ var findSeiNals = function(avcStream, samples, trackId) {
         seiNal.pts = lastMatchedSample.pts;
         seiNal.dts = lastMatchedSample.dts;
       } else {
-        // eslint-disable-next-line no-console
-        console.log("We've encountered a nal unit without data. See mux.js#233.");
+        logWarn("We've encountered a nal unit without data. See mux.js#223.");
```
brandonocasey, line 101: The issue number was wrong here. It was actually 223 not 233

gkatsev (approved):

lib/m2ts/metadata-stream.js
```diff
@@ -91,8 +91,6 @@ var
 MetadataStream = function(options) {
   var
     settings = {
-      debug: !!(options && options.debug),
```
brandonocasey, line 94: debug was only used for this console.log, now we just trigger an event at any time and the user can choose what to do.

gkatsev (approved):

PR #393, opened 2021-07-13
fix: Do not scale width by sarRatio, we already generate a pasp box with sarRatio information.

brandonocasey:
While this doesn't give us the absolute width when we have a `sarRatio`, we don't want to use the absolute width. This is because we already encode the `sarRatio` in the `pasp` atom of our mp4 output. By encoding the `sarRatio` and an absolute width, the width is skewed even further unless the decoder looks directly at the nal units and calculates it itself. Example:
* `sarRatio` is `0.5`
* `width` is `640` without `sarRatio`

How it is now with `sarRatio` width would get set to `320` and then when the decoder see's the `pasp` atom with `sarRatio` it will think that width is `160`. Some decoders will ignore all these values on use nal units to calculate the width, but mobile Samsung Chrome on Android 11 will not. This causes a media decode error as the actual width of the video is `320`. 

With this change width will be set to `640` and the `sarRatio` `pasp` encoding will make the decoder set the width to `320`

### DIFF
--- a/lib/codecs/h264.js
+++ b/lib/codecs/h264.js
@@ -151,7 +151,8 @@ PROFILES_WITH_OPTIONAL_SPS_DATA = {
   118: true,
   128: true,
 
-  // TODO: these three don't appear to be in the specificiation anymore?
+  // TODO: the three profiles below don't
+  // appear to have sps data in the specificiation anymore?
   138: true,
   139: true,
   134: true
--- a/lib/codecs/h264.js
+++ b/lib/codecs/h264.js
@@ -214,10 +214,10 @@ H264Stream = function() {
         pts: currentPts,
         dts: currentDts,
         data: data,
-        nalUnitType: data[0] & 0x1f
+        nalUnitTypeCode: data[0] & 0x1f
       };
 
-    switch (event.nalUnitType) {
+    switch (event.nalUnitTypeCode) {
     case 0x05:
       event.nalUnitType = 'slice_layer_without_partitioning_rbsp_idr';
       break;
--- a/lib/codecs/h264.js
+++ b/lib/codecs/h264.js
@@ -150,6 +150,8 @@ PROFILES_WITH_OPTIONAL_SPS_DATA = {
   86: true,
   118: true,
   128: true,
+
+  // TODO: these three don't appear to be in the specificiation anymore?
   138: true,
   139: true,
   134: true
@@ -365,7 +367,7 @@ H264Stream = function() {
       picHeightInMapUnitsMinus1,
       frameMbsOnlyFlag,
       scalingListCount,
-      sarRatio,
+      sarRatio = [1, 1],
       aspectRatioIdc,
       i;
 
--- a/lib/codecs/h264.js
+++ b/lib/codecs/h264.js
@@ -210,10 +210,11 @@ H264Stream = function() {
         trackId: trackId,
         pts: currentPts,
         dts: currentDts,
-        data: data
+        data: data,
+        nalUnitType: data[0] & 0x1f
       };
 
-    switch (data[0] & 0x1f) {
+    switch (event.nalUnitType) {
     case 0x05:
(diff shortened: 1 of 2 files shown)


lib/codecs/h264.js
```diff
@@ -210,10 +210,11 @@ H264Stream = function() {
         trackId: trackId,
         pts: currentPts,
         dts: currentDts,
-        data: data
+        data: data,
+        nalUnitType: data[0] & 0x1f
       };
 
-    switch (data[0] & 0x1f) {
+    switch (event.nalUnitType) {
```
brandonocasey, line 217: It's annoying to try and determine the type of nal unit if it's not one that we normally parse.

lib/codecs/h264.js
```diff
@@ -470,7 +470,7 @@ H264Stream = function() {
       profileIdc: profileIdc,
       levelIdc: levelIdc,
       profileCompatibility: profileCompatibility,
-      width: Math.ceil((((picWidthInMbsMinus1 + 1) * 16) - frameCropLeftOffset * 2 - frameCropRightOffset * 2) * sarScale),
+      width: (((picWidthInMbsMinus1 + 1) * 16) - frameCropLeftOffset * 2 - frameCropRightOffset * 2),
```
brandonocasey, line 473: We shouldn't need to Math.ceil if we are not scaling now.

lib/codecs/h264.js
```diff
@@ -150,6 +150,8 @@ PROFILES_WITH_OPTIONAL_SPS_DATA = {
   86: true,
   118: true,
   128: true,
+
+  // TODO: these three don't appear to be in the specificiation anymore?
```
brandonocasey, line 154: These don't appear in the spec, but I doubt there is any harm to keeping them here for now.

lib/codecs/h264.js
```diff
@@ -365,7 +367,7 @@ H264Stream = function() {
       picHeightInMapUnitsMinus1,
       frameMbsOnlyFlag,
       scalingListCount,
-      sarRatio,
+      sarRatio = [1, 1],
```
brandonocasey, line 370: default to a 1 to 1 aspect ratio, ie width and height are exact values.

lib/codecs/h264.js
brandonocasey: I think the original reason that we did scaling is because we did not encode a `pasp` atom with sarRatio information when mux.js was created. So We had to encode width and height as their real values at all times. I think we should have been scaling height in some scenarios too. Overall we won't have to do any of that now as the decoder will see the `pasp` and scale accordingly. We just have to report the values that exist.

lib/codecs/h264.js
gkatsev: which three? 138,139,134?

lib/codecs/h264.js
gkatsev: so now, for like sei_rbsp, the nalUnitType would end up being `sei_rbsp` but others would just pass along the hex value of it?

lib/codecs/h264.js
brandonocasey: yeah the three below don't appear at the lines specified in the specs that we have.

brandonocasey: updated the comment to clarify it a bit better.

lib/codecs/h264.js
brandonocasey: Yeah, thinking about it again I think I will change it to `nalUnitTypeCode` that way type comparisons won't break down the line, since we expect it to not exist or be a string.

lib/codecs/h264.js
gkatsev: Just noticed that diff hid the next line that would've shown there isn't more. I guess I forgot to expand it and that made it a bit unclear. The new comment is nicer, though, thanks!

gkatsev (approved):

PR #394, opened 2021-08-17
feat: add option to specify starting sequence number for Transmuxer

collin:
We have a service that generates ISO BMFF chunks for consumption with MSE.

In order to process only the chunks requested we don't always want start our sequence numbers at zero.

Adding an option to specify a starting sequence number gives us the necessary control to create arbitrary chunks in a longer logical sequence of media segments.

This PR includes build artifacts because we're using the branch as our source to install, if this change is acceptable I can re-create a clean PR without the build artifacts included.

I am also curious to know if this is the best way to test this kind of change, I've copied one of the other Transmuxer tests as a starting point, but I'm not sure if everything in the test I copied is neccessary.

I considered also testing the VideoSegmentStream/AudioSegmentStream individually. If that's desirable I can create those tests.


gkatsev:
Hi @collin this seems reasonable to add, and the actual change is pretty small. If you're able to clean-up the PR (without the dist/ files) that would be swell.
I've asked my teammates to look at this and see whether there are any issues with the way the test is.

test/transmuxer.test.js
```diff
@@ -3427,6 +3427,59 @@ QUnit.test('no options creates combined output', function(assert) {
...
+
+  boxes = mp4.tools.inspect(segments[0].data);
+  moof = boxes.find(function(box) {
+    return box.type === 'moof';
+  });
+  mfhd = moof.boxes.find(function(box) {
+    return box.type === 'mfhd';
+  });
+
+  assert.equal(mfhd.sequenceNumber, 10, 'requested first sequence number starts at 10');
```
brandonocasey, line 3480: We should check both generated mfhd's for the sequence number here. Should be something like 
```js
  mfhds = [];
  boxes = mp4.tools.inspect(segments[0].data);

  boxes.forEach(function(box) {
    if (box.type === 'moof') {
      box.boxes.forEach(function(_box) {
        if (_box.type === 'mfhd') {
          mfhds.push(_box);
        }
      });
    }
  });

  assert.equal(mfhds.length, 2, 'two mfhds');

  mfhds.forEach(function(mfhd) {
    assert.equal(mfhd.sequenceNumber, 10, 'requested first sequence number starts at 10');
  });

```

brandonocasey (changes requested):

collin:
@gkatsev @brandonocasey  #395 is a cleaner branch.

PR #395, opened 2021-08-24
Adds mfhd starting sequence option to Transmuxer

collin:
This PR replaces #394. It removes build artifacts and address the change in the test suggested by @brandonocasey.

/cc @gkatsev 

Adds an option `firstSequenceNumber` to Transmuxer.

Can be used like this:

```js
new Transmuxer({ firstSequenceNumber: 10 });
```

This is passed in to VideoSegmentStream and AudioSegmentStream.

When mfhd boxes are constructed, the starting sequence number will be
taken from that option. The default value is zero.

This is useful for generating segmented streams out of order. Out of
oder processing allows for restart, parallelism, or random seeking into
source material.

### DIFF
--- a/lib/mp4/transmuxer.js
+++ b/lib/mp4/transmuxer.js
@@ -92,12 +92,14 @@ var generateSegmentTimingInfo = function(
 AudioSegmentStream = function(track, options) {
   var
     adtsFrames = [],
-    sequenceNumber = 0,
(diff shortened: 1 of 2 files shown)


brandonocasey (approved):

brandonocasey:
This is out in 5.13.0 Thanks @collin

PR #398, opened 2021-08-31
feat: Add multibyte character support

alex-barstow:
## Description
This adds multibyte character support to the 708 implementation via a new `encoding` option that can be passed to the `CaptionStream` constructor. The intent is to permit legacy multi-byte encodings for Chinese, Japanese, Korean, and others. The value of the option is used to create a [`TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder) instance (assuming the current browser supports it, only IE doesn't), so the value can be any ["valid [encoding] label"](https://developer.mozilla.org/en-US/docs/Web/API/Encoding_API/Encodings).

I'm waiting to commit the test I wrote until I have an open-source content sample to use


lib/m2ts/caption-stream.js
```diff
@@ -610,8 +638,23 @@ Cea708Stream.prototype.getPts = function(byteIndex) {
...
+  // Check if an encoding value was specified for this service
+  for (var i = 0; i < serviceNames.length; i++) {
+    serviceName = serviceNames[i];
+    providedEncoding = this.serviceEncodings[serviceName];
+
+    if (serviceName === 'SERVICE' + serviceNum) {
+      encoding = providedEncoding;
+      break;
+    }
+  }
```
gkatsev, line 655: since `serviceEncodings` is a mapping of service names to encodings, you can use the `in` operator, to know whether the object has that property or not. So, we should be able to simplify this to be the following
```suggestion
  var serviceName = 'SERVICE' + serviceNum;
  if (serviceName in this.serviceEncodings) {
    encoding = this.serviceEncodings[serviceName];
  }
```

lib/m2ts/caption-stream.js
```diff
@@ -628,14 +671,35 @@ Cea708Stream.prototype.initService = function(serviceNum, i) {
...
+
+      // Skip the next byte in the loop since it has already been handled
+      i++;
+    } else {
+     charCodeArray = [currentByte];
+    }
+
+    char = service.textDecoder_.decode(new Uint8Array(charCodeArray));
+  } else {
+    char = get708CharFromCode(extended | currentByte);
```
gkatsev, line 696: so, `get708CharFromCode` method seems to account for some extended characters. We probably need to do the same for the encoded stuff too.
I wonder whether moving the text decoder piece into `flushDisplayed` would work. The benefit is that it could decode once per cue rather than once per character. Unfortunately, didn't get a chance to try it out.

lib/m2ts/caption-stream.js
alex-barstow: Much better

lib/m2ts/caption-stream.js
```diff
@@ -406,10 +411,45 @@ Cea708Service.prototype.setCurrentWindow = function(windowNum) {
...
+ */
+Cea708Service.prototype.createTextDecoder = function(encoding) {
+  if (typeof TextDecoder === 'undefined') {
+    this.trigger('log', {
+      level: 'warn',
+      message: 'The `encoding` option is unsupported without TextDecoder support'
+    });
+  } else {
+    try {
+      this.textDecoder_ = new TextDecoder(encoding);
```
brandonocasey, line 425: This appears to throw if the encoding was incorrect. We may want to concat the caught error onto our warning below.

lib/m2ts/caption-stream.js
```diff
@@ -626,22 +677,62 @@ Cea708Stream.prototype.initService = function(serviceNum, i) {
...
+  var currentByte = packetData[i];
+  var nextByte = packetData[i + 1];
   var win = service.currentWindow;
+  var char;
+  var charCodeArray;
+
+  // Use the TextDecoder if one was created for this service
+  if (service.textDecoder_ && !isExtended) {
+    if (isMultiByte) {
+      charCodeArray = [currentByte, nextByte];
```
brandonocasey, line 694: We should subarray here instead to prevent the creation of a new Uint8Array below. 
```
// for multi-byte (can they be more than two characters?)
charCodeArray = packetData.subarray(i, i + 1)

// for single
charCodeArray = packetData.subarray(i)
```

brandonocasey (commented):
Could use a test for:
* passing the encoding around
* Multi-byte character examples (if we can pull generic words out of our sample for use)
* Not having a TextDecoder
* An invalid encoding

lib/m2ts/caption-stream.js
alex-barstow: Good call

lib/m2ts/caption-stream.js
```diff
@@ -406,10 +411,45 @@ Cea708Service.prototype.setCurrentWindow = function(windowNum) {
...
+    } catch (error) {
+      this.trigger('log', {
+        level: 'warn',
+        message: 'TextDecoder could not be created for provided `encoding` value: ' + encoding
+      });
+    }
+  }
+}
+
+var Cea708Stream = function(options) {
```
gkatsev, line 435: we should add a null check for options, should help with a lot of the test failures
```suggestion
var Cea708Stream = function(options) {
  var options = options || {};
```

lib/m2ts/caption-stream.js
alex-barstow: @brandonocasey Actually, `packetData` isn't a typed array so we do need to create a Uint8Array here. And to address your other question:
> When a decoder encounters the P16 code, it uses the succeeding **two bytes** to
address characters in a 16-bit code set.

lib/m2ts/caption-stream.js
alex-barstow: I fixed the flow to only use a `TextDecoder` if a `P16` code is encountered (indicating a multibyte character), but to still use `get708CharFromCode()` when an `EXT1` code is encountered (indicating a character from an extended code set). This should address your concern.

brandonocasey (approved):
Approved pending the tests pass

test/caption-stream.test.js
```diff
@@ -2735,6 +2761,69 @@ QUnit.test('parses 708 captions', function(assert) {
...
+        encoding: 'invalid'
+      }
+    }
+  });
+
+  cc708Korean.forEach(secondCea708Stream.push, secondCea708Stream);
+  secondCea708Stream.flushDisplayed(4721138662, secondCea708Stream.services[1]);
+
+  assert.notOk(secondCea708Stream.services[1].textDecoder_, 'TextDecoder not created when encoding is invalid');
+});
```
gkatsev, line 2825: this test needs to check for IE11 or if TextDecoder is present.

alex-barstow: Added a check

gkatsev (approved):
Verified locally. Can see captions displayed properly instead of random characters!

PR #408, opened 2022-02-18
Send ID3 tag even when it has malformed content

pszemus:
Enables sending ID3 tag with as many frames as could be parsed.

Without this change [this.trigger('data', tag)](https://github.com/videojs/mux.js/blob/main/lib/m2ts/metadata-stream.js#L249) wasn't called if stumbled upon malformed ID3 tag/frame.

Beneficiaries: Tellos Z/IP Stream encoder produces ID3 data with `TIT2`, `WOAF` and `TENC` frames, from which only `TIT2` can be parsed by mux.js, so let's take those frames that have been parsed.

Additionally, I'll take a look at mux.js' ID3 parser and try to find out if that metadata is really malformed or it's an issue with the parser.

### DIFF
--- a/lib/m2ts/metadata-stream.js
+++ b/lib/m2ts/metadata-stream.js
@@ -200,7 +200,7 @@ MetadataStream = function(options) {
           level: 'warn',
           message: 'Malformed ID3 frame encountered. Skipping metadata parsing.'
         });
-        return;
+        break;
       }
       frameHeader = String.fromCharCode(tag.data[frameStart],
                                         tag.data[frameStart + 1],
--- a/lib/m2ts/metadata-stream.js
+++ b/lib/m2ts/metadata-stream.js
@@ -198,8 +198,10 @@ MetadataStream = function(options) {
       if (frameSize < 1) {
         this.trigger('log', {
           level: 'warn',
-          message: 'Malformed ID3 frame encountered. Skipping metadata parsing.'
+          message: 'Malformed ID3 frame encountered. Skipping remaining metadata parsing.'
         });
+        // If the frame is malformed, don't parse any further frames but allow previous valid parsed frames
+        // to be sent along.
         break;
       }
       frameHeader = String.fromCharCode(tag.data[frameStart],
--- a/test/metadata-stream.test.js
+++ b/test/metadata-stream.test.js
@@ -518,6 +518,27 @@ QUnit.test('constructs the dispatch type', function(assert) {
   assert.equal(metadataStream.dispatchType, '1503020100', 'built the dispatch type');
 });
 
+QUnit.test('should skip tag frame parsing on malformed frame, preserving previous frames', function(assert) {
+  var events = [],
+      validFrame = id3Frame('TIT2',
+                            0x03, // utf-8
+                            stringToCString('sample title')),
+      malformedFrame = id3Frame('WOAF'), // frame with size of 0B
+      tag = id3Tag(validFrame, malformedFrame);
+  
+  metadataStream.on('data', function(event) {
+    events.push(event);
+  });
+
+  metadataStream.push({
+    type: 'timed-metadata',
+    data: new Uint8Array(tag)
+  })
+  
+  assert.equal(events.length, 1, 'parsed 1 tag')
+  assert.equal(events[0].frames.length, 1, 'parsed one frame');
+  assert.equal(events[0].frames[0].key, 'TIT2');
+});
 
 QUnit.test('can parse PRIV frames in web worker', function(assert) {
   var payload = stringToInts('arbitrary'),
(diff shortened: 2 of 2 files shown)


pszemus:
can anyone pick up this PR?

lib/m2ts/metadata-stream.js
```diff
@@ -200,7 +200,7 @@ MetadataStream = function(options) {
           level: 'warn',
           message: 'Malformed ID3 frame encountered. Skipping metadata parsing.'
```
gesinger, line 201:
```suggestion
          message: 'Malformed ID3 frame encountered. Skipping remaining metadata parsing.'
```

lib/m2ts/metadata-stream.js
```diff
@@ -200,7 +200,7 @@ MetadataStream = function(options) {
...
         });
-        return;
+        break;
```
gesinger, line 203:
```suggestion
        // If the frame is malformed, don't parse any further frames but allow previous valid parsed frames
        // to be sent along.
        break;
```

gesinger (commented):
Thank you for the PR!

I think this change makes sense as a more lenient approach to malformed content, but it may be worth looking at the encoder and content themselves.

If this is the line that breaks the sample content, then the sample content appears to not have a valid frame header, as the size should be the size of the frame (which should be greater than 0).

Judging from https://github.com/videojs/mux.js/blob/1838b7f62c729c239a325a0c65036c4812215bd9/lib/m2ts/metadata-stream.js#L216 and https://github.com/videojs/mux.js/blob/1838b7f62c729c239a325a0c65036c4812215bd9/lib/m2ts/metadata-stream.js#L40-L88 it seems that mux.js supports TXXX, WXXX, and PRIV, and will break out of frame parsing for any other types.

Did you have a specific segment that's breaking on this line? I'd be interested to take a look, and if you have any other details on the ID3 tags that might help.

It would also be worth us adding a test, if you don't mind writing one up.

Thank you again!

pszemus:
> it may be worth looking at the encoder and content themselves

After further investigation it turns out that, if the encoder is missing the audio URL, it creates an ID3 `WOAF` frame with size of 0B, which is not allowed according to ID3 documentation. Encoder should insert an ID3 frame with empty string or abort inserting such a frame. I've managed to report that case to the developer.

Nevertheless I'd recommend merging this PR.

> it seems that mux.js supports TXXX, WXXX, and PRIV, and will break out of frame parsing for any other types.

mux.js won't decode the information sent in text frames other than `TXXX` (e.g. `TIT2` the encoder is using for holding the song title), but it passes the raw binary data to the client (`frame.data`)
https://github.com/videojs/mux.js/blob/1838b7f62c729c239a325a0c65036c4812215bd9/lib/m2ts/metadata-stream.js#L213
so I'm able to decode it myself.

I'll be willing to spend some more time and try to implement decoding of text content for all text frames (`T*`, e.g. `TIT2`).

> Did you have a specific segment that's breaking on this line? I'd be interested to take a look

I have a 5sec .ts file with malformed ID3 frame in a way described in the beginning of this commend. I can send it to you if you're interested.

> It would also be worth us adding a test, if you don't mind writing one up.

I'll try my best and get back to you soon.

test/metadata-stream.test.js
```diff
@@ -518,6 +518,27 @@ QUnit.test('constructs the dispatch type', function(assert) {
   assert.equal(metadataStream.dispatchType, '1503020100', 'built the dispatch type');
 });
 
+QUnit.test('should skip tag frame parsing on malformed frame, preserving previous frames', function(assert) {
+  var events = [],
+      validFrame = id3Frame('TIT2',
+                            0x03, // utf-8
+                            stringToCString('sample title')),
+      malformedFrame = id3Frame('WOAF'), // frame with size of 0B
```
pszemus, line 526: I could also create the same malformed frame like this:
```
stringToInts('WOAF').concat([
        0x00, 0x00, 0x00, 0x00, // size
        0x00, 0x00 // flags
      ]),
```
to make the malformation more visible

gesinger (approved):

gesinger:
Thank you again @pszemus , I think you're right that we should add this despite it being a content issue. If encoders take the approach of including malformed ID3 frames it still makes sense to pass back valid ones already parsed.

> I'll be willing to spend some more time and try to implement decoding of text content for all text frames (T*, e.g. TIT2).

Any contributions are very much appreciated if you have the time!

> I have a 5sec .ts file with malformed ID3 frame in a way described in the beginning of this commend. I can send it to you if you're interested.

If the content is not public enough to share here, and you can share it via another medium (e.g., the [Video.js Slack](http://slack.videojs.com)), let me know. It would be good for testing.

And thank you for adding the test!

pszemus:
This is the .ts file with malformed `WOAF` frame (0B content size): [seg64000-00031954.ts.zip](https://github.com/videojs/mux.js/files/8779741/seg64000-00031954.ts.zip)

If you're happy with this PR you can merge it. Changes regarding parsing other frames will be sent in another PR.

misteroneill (approved):

gkatsev (approved):

PR #417, opened 2022-07-05
fix: replace indexOf with typedArrayIndexOf for IE11 support

gesinger:


gkatsev (approved):

pszemus (approved):
@gesinger You could also replace for loop with `typedArrayIndexOf` for `PRIV` frames.

pszemus (approved):

PR #426, opened 2022-12-29
feat: support emsg box parsing

adrums86:
**Feature**
--
Added basic fmp4 [emsg box](https://dashif-documents.azurewebsites.net/Events/master/event.html#emsg-format) parsing for v0 and v1 boxes. Exposed a new `mp4/probe` function for extracting timed message data from emsg boxes `getEmsgID3`. 
This will  allow us to extract emsg timed metadata from fmp4/cmaf content, specifically ID3 frame data which is supported by various ad service integrations.

References:
https://aomediacodec.github.io/id3-emsg/
https://dashif-documents.azurewebsites.net/Events/master/event.html#event-metadata-timing
https://github.com/id3/ID3v2.4/blob/master/id3v2.40-structure.txt

**Testing**
--
Tested using mock data derived from the emsg spec, as well as manually tested [real fmp4 segment data](https://storage.googleapis.com/dlevin/tmp/fmp4-emsg-id3/1567191409/segment-0-1.mp4) in the debug page.


lib/mp4/probe.js
```diff
@@ -370,6 +371,30 @@ getTracks = function(init) {
...
+                          parsedBox.presentation_time_delta, 
+                          ptOffset);
+      var duration = emsg.scaleTime(parsedBox.event_duration, parsedBox.timescale);
+      emsgID3Data.push({
+          pts,
+          duration,
+          data: parsedBox.message_data,
+      });
+  });
+  return emsgID3Data;
```
dzianis-dashkevich, line 395: We can simplify with a map:
`return emsgBoxes.map((data) => { ... return { pts, duration, data }})`

dzianis-dashkevich (approved):

lib/mp4/probe.js
adrums86: Good call on using `map`, I often operate in the mode of trying to write code that "fits" into the codebase, rather than using the most suitable syntax for the task.

lib/utils/string.js
```diff
@@ -0,0 +1,20 @@
...
+    var retString = '';
+    while (curChar !== '\0') {
+        retString += curChar;
+        index++;
+        curChar = String.fromCharCode(data[index]);
+    }
+    // Add nullChar
+    retString += curChar;
+    return retString;
+  };
```
wseymour15, line 18: nit: It looks like this end bracket is not lined up with the function declaration. You also may want to align the module.export statement with the function as well.

test/mp4-emsg.test.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
wseymour15, line 1: Not a huge deal either, but it looks like most of the other files are indented with 2 spaces and this one is 4.

test/utils.string.test.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
```
wseymour15, line 1: Same with this file, I'd probably keep this one indented with 2 spaces consistently.

wseymour15 (changes requested):
This looks great 👍 Parsing by bytes looks painful.. I'm sure that took some effort to figure out! +1 after some minor styling changes and possibly another test or two.

adrums86:
@wseymour15 good catch on the indentation inconsistencies, I was leaning on auto-detect in the IDE without realizing my default was 4 per tab 😅 . I added a couple more edge case tests for the c string util as well. Thanks for taking a look at this!

wseymour15 (approved):

test/mp4-emsg.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+QUnit.test('Can parse a v0 emsg box', function(assert) {
+  var boxData = generateEmsgBoxData(0, messageData);
+  var parsedBox = emsg.parseEmsgBox(boxData);
+
+  assert.equal(parsedBox.scheme_id_uri, 'urn:foo:bar:2023\0', 'v0 box has expected scheme_id_uri');
+  assert.equal(parsedBox.value, 'foo.bar.value\0', 'v0 box has expected value');
+  assert.equal(parsedBox.timescale, 100, 'v0 box has expected timescale');
+  assert.equal(parsedBox.presentation_time, undefined, 'v0 box has expected presentation_time');
+  assert.equal(parsedBox.presentation_time_delta, 1000, 'v0 box has expected presentation_time_delta');
```
wseymour15, line 18: No change required but out of curiosity, why is the presentation_time value undefined here, and the presentation_time_delta is undefined in the other test?

misteroneill (approved):

lib/tools/parse-id3.js
```diff
@@ -0,0 +1,250 @@
...
+
+// TODO add dispatch type parsing
+var parseId3Frames = function(data) {
+  var frameSize, frameHeader,
+    frameStart = 10,
+    tagSize = 0,
+    frames = [];
+
+  // if not valid id3 return early.
+  if (data.length < 10 ||
```
wseymour15, line 184: It might be useful here to give a bit more of an explanation to why it isn't valid. Specifically why the length can't be less than 10.

lib/tools/parse-id3.js
```diff
@@ -0,0 +1,250 @@
...
+  // the frame size is transmitted as a 28-bit integer in the
+  // last four bytes of the ID3 header.
+  // The most significant bit of each byte is dropped and the
+  // results concatenated to recover the actual value.
+  tagSize = parseSyncSafeInteger(data.subarray(6, 10));
+
+  // ID3 reports the tag size excluding the header but it's more
+  // convenient for our comparisons to include it
+  tagSize += 10;
+  if (data[5] & 0x40) {
```
wseymour15, line 199: It may also be useful to assign `data[5]` to a more descriptive variable here.

test/mp4-emsg.test.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+var QUnit = require('qunit'),
```
wseymour15, line 3: Would it be useful to also have a test or two for invalid data?

lib/tools/parse-id3.js
dzianis-dashkevich: I assume this is a bitmask for flags to check if an extended header is present:
https://github.com/id3/ID3v2.4/blob/master/id3v2.40-structure.txt

lib/tools/parse-id3.js
dzianis-dashkevich: I guess this is header length in case if data is Uint8

test/mp4-emsg.test.js
adrums86: `presentation_time` and `presentation_time_delta` usage is dependent on the emsg box version. `v0` boxes use `presentation_time_delta` and `v1` boxes use `presentation_time`. I'm not exactly sure as to _why_ they are that way, but it does change the calculation for the cueTime, or the `ST` from the [spec](https://dashif-documents.azurewebsites.net/Events/master/event.html#emsg-format).

test/mp4-emsg.test.js
adrums86: Good call, I'll throw together a few new tests with malformed data.

lib/tools/parse-id3.js
adrums86: Yeah, this is the extended header bitmask. I'll add a comment and/or assign this to something more descriptive.

lib/tools/parse-id3.js
adrums86: I'll add a bit more of an explanation here.

brandonocasey (commented):
I think we have some existing code (now duplicate) for this in https://github.com/videojs/mux.js/tree/main/lib/aac/utils.js and https://github.com/videojs/mux.js/tree/main/lib/aac/index.js

adrums86:
@brandonocasey Yeah, I noticed there was already some duplication between the aac timed metadata parsing and `m2ts/metadata-stream`. https://github.com/videojs/mux.js/blob/2e507042ef3d9932c0befb5f3de721db5bb095a7/lib/m2ts/metadata-stream.js#L206
I lifted as much common functionality from `metadata-stream` as I could, however since both aac and m2ts are accommodating partial segments, it seemed safer to leave the `push` implementations alone and implement extraction from full-segment fmp4 separately. There is likely room for further consolidation between all 3 implementations, however I think it makes sense to do that in a separate PR.

adrums86:
@wseymour15 @dzianis-dashkevich I added some emsg validation and corresponding tests to account for malformed data and addressed the other comments. Thanks for taking a look at the revised PR!

PR #428, opened 2023-02-20
fix: emsg ie11 test failures

adrums86:
Use the number utils `getUint64` function and change uses of `Array.from` to `[].slice.call` in unit tests for IE11 support.


Essk (approved):
1655296268402_e=2147483647 v=beta t=DGGcdCjf0oDinfevtE4rFOw-KoztErK99Rpd8I8BVdc

PR #430, opened 2023-05-09
fix(aac): missing timestamp rollover

amtins:
## Description
This PR fixes the missing timestamp rollover in AAC stream packets.

Refer https://github.com/videojs/http-streaming/issues/1371#issue-1582588141

## Specific Changes proposed

- handle PTS and DTS  timestamp rollover

## Requirements Checklist
- [X] Feature implemented / Bug fixed
- [ ] If necessary, more likely in a feature request than a bug fix
  - [ ] Change has been verified in an actual browser (Chrome, Firefox, IE)
  - [ ] Unit Tests updated or fixed
  - [ ] Docs/guides updated
  - [ ] Example created ([starter template on JSBin](https://codepen.io/gkatsev/pen/GwZegv?editors=1000#0))

### DIFF
--- a/lib/aac/index.js
+++ b/lib/aac/index.js
@@ -11,6 +11,8 @@
 'use strict';
 var Stream = require('../utils/stream.js');
 var aacUtils = require('./utils');
+var handleRollover =
+  require('../m2ts/timestamp-rollover-stream.js').handleRollover;
 
 // Constants
 var AacStream;
@@ -22,7 +24,8 @@ var AacStream;
 AacStream = function() {
   var
     everything = new Uint8Array(),
-    timeStamp = 0;
+    timeStamp = 0,
+    referenceDTS;
 
   AacStream.prototype.init.call(this);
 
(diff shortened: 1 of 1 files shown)


amtins:
@adrums86 thank you for taking the time to review this PR.

Unfortunately, I'm not sure I'm much help with this PR, it was only by luck that I managed to find a fix. So it is quite possible that there is a better way to fix this behavior.

I would have liked to add unit tests, but unfortunately I can't figure out how it all works so I can't produce a test case. 

However, if you need a media to test I have a sample that I can share via slack.

adrums86:
@amtins, no worries, your contribution is very much appreciated! I'll spend some time in the next week or so and see if I can help at all with a test or possible segment extraction of the `dts`.

adrums86:
I think this is a great improvement in general and we should roll forward with this change if possible. I'd like to get another set of eyes on this, however. @dzianis-dashkevich can you take a look at this when you have a chance and possibly weigh in on my original question, regarding extracting the AAC `dts` value in a different way or does this seem the best way to go about it? TIA.

amtins:
@adrums86 I tried to add test coverage as best I could.

adrums86:
@amtins Thank you for getting back to this and adding a test! This LGTM.

dzianis-dashkevich (approved):

dzianis-dashkevich:
@adrums86 , I believe that we have only encoded audio data on the AAC level without timing or synchronization information. I think pts/dts info is only available on the container level (ts/mp4)

adrums86:
@dzianis-dashkevich makes sense. Big +1 to moving to CMAF 😄 . Lets merge this thing. Thanks again for doing the work @amtins!

Ruud-Gerrits:
Hi, any idea on when this will be merged?

PR #434, opened 2023-07-11
feat: add position data to captions

wseymour15:
NOTE: This will require a major version update, as this changes how the caption data is sent to VHS. This was necessary, as we were previously combining text that can have different positioning data on these messages. This will require changes in VHS and video.js once this is released.

Previously for CEA-608 captions, we were completely ignoring positioning data and just centering all captions at the bottom of the screen. This change allows us to take advantage of our parsing and including this positioning data. The 'line' field represents the row on which to display the text from 1-15, and the `position` field represents where the caption should be displayed horizontally on a scale from 1-100 (although the spec states we should lock it between 10-80).
See: https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-608

Testing instructions will be included once I have a PR out with the VHS changes to consume this change.

### DIFF
--- a/lib/flv/coalesce-stream.js
+++ b/lib/flv/coalesce-stream.js
@@ -35,7 +35,7 @@ var CoalesceStream = function(options) {
   this.push = function(output) {
     // buffer incoming captions until the associated video segment
     // finishes
-    if (output.text) {
+    if (output.content) {
       return this.pendingCaptions.push(output);
     }
     // buffer incoming id3 tags until the final flush
--- a/lib/flv/coalesce-stream.js
+++ b/lib/flv/coalesce-stream.js
@@ -35,7 +35,7 @@ var CoalesceStream = function(options) {
   this.push = function(output) {
     // buffer incoming captions until the associated video segment
     // finishes
-    if (output.content) {
+    if (output.content || output.text) {
       return this.pendingCaptions.push(output);
     }
     // buffer incoming id3 tags until the final flush
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -1233,7 +1233,7 @@ var ROWS = [0x1100, 0x1120, 0x1200, 0x1220, 0x1500, 0x1520, 0x1600, 0x1620,
 
 // CEA-608 captions are rendered onto a 34x15 matrix of character
 // cells. The "bottom" row is the last element in the outer array.
-// We keep track of positioning infomration as we go by storing the
+// We keep track of positioning information as we go by storing the
 // number of indentations and the tab offset in this buffer.
 var createDisplayBuffer = function() {
   var result = [], i = BOTTOM_ROW + 1;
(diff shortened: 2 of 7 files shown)


lib/flv/coalesce-stream.js
```diff
@@ -35,7 +35,7 @@ var CoalesceStream = function(options) {
   this.push = function(output) {
     // buffer incoming captions until the associated video segment
     // finishes
-    if (output.text) {
+    if (output.content || output.text) {
```
wseymour15, line 38: I believe other types of captions (708) could still use text as of now, so both content and text are included.

lib/m2ts/caption-stream.js
```diff
@@ -1427,7 +1437,11 @@ var Cea608Stream = function(field, dataChannel) {
         // increments the column cursor by 4, so we can get the desired
         // column position by bit-shifting to the right (to get n/2)
         // and multiplying by 4.
-        this.column_ = ((data & 0xe) >> 1) * 4;
+        const indentations = ((data & 0xe) >> 1);
+
+        this.column_ = indentations * 4;
+        // add to the number of indentations for positioning
+        this.nonDisplayed_[this.row_].indent += indentations;
```
adrums86, line 1444: Do we need to add the indent to the `this.displayed_` buffer here as well?

lib/m2ts/caption-stream.js
```diff
@@ -1390,7 +1394,13 @@ var Cea608Stream = function(field, dataChannel) {
       // increments, with an additional offset code of 1-3 to reach any
       // of the 32 columns specified by CEA-608. So all we need to do
       // here is increment the column cursor by the given offset.
-      this.column_ += (char1 & 0x03);
+      const offset = (char1 & 0x03);
+
+      // For an offest value 1-3, set the offset for that caption
+      // in the non-displayed array.
+      this.nonDisplayed_[this.row_].offset = offset;
```
adrums86, line 1401: Do we need to add the offset to the `this.displayed_` buffer?

adrums86 (commented):
A couple questions about applying the `offset` and `indent` to `this.displayed_`. Also have we tried this with 708? Curious if we get the positioning there as well, I can dig up some media or maybe we can create a unit test to try it out. Overall this looks great though, nice work!

lib/m2ts/caption-stream.js
wseymour15: Good question, and I don't believe so. I think the `this.nonDisplayed_` buffer is basically where the caption data is manipulated until the end of a caption. At that instance, [we swap everything over to `this.displayed_`](https://github.com/videojs/mux.js/blob/dc56f1b79224bf20dc34ec1e949df6b18c2ffde4/lib/m2ts/caption-stream.js#L1290).

wseymour15:
> Also have we tried this with 708?

@adrums86 Great question! As of now, these changes will not affect the position of 708 captions. I do however think we could add this logic to the `flushDisplayed` function of the 708 captions [following the spec](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-708). It looks like we have some of the backbones for determining things like priority and window anchor that the spec needs, so this is definitely doable.

I will add tickets on the Brightcove side to address this work.

adrums86 (approved):
Great work!

PR #437, opened 2023-09-07
fix(m2ts): PES packets missing timestamp rollover

amtins:
## Description

Handles `PTS` and `DTS` titmestamp rollover separately, directly in PES packets.


wseymour15:
Can we close this PR? It appears that it was handled in https://github.com/videojs/http-streaming/pull/1426

PR #439, opened 2023-10-11
fix: 708 captions multi-byte char fix

wseymour15:
We were seeing issues with non-latin 708 captions not being displayed correctly. It appears that the issue that we are not always handling multi-byte characters correctly.

This fix proposes that in the case where we see a multi-byte character without a pre-designated decoder, we send the byte array through a function to turn it into a hex string, and then from there decode it into the character.

Example Test Playlist: https://solutions.brightcove.com/wseymour/708-testing/playlist.m3u8

Before: 
Screenshot 2023-10-12 at 10 47 59 AM

After:
Screenshot 2023-10-12 at 10 46 12 AM

### DIFF
--- a/lib/m2ts/caption-stream.js
+++ b/lib/m2ts/caption-stream.js
@@ -690,9 +690,12 @@ Cea708Stream.prototype.handleText = function(i, service, options) {
 
   // Converts an array of bytes to a unicode hex string.
   function toHexString(byteArray) {
-    return Array.from(byteArray, function(byte) {
-      return ('0' + (byte & 0xFF).toString(16)).slice(-2);
-    }).join('');
+    const newArr = [];
+    byteArray.forEach((byte) => {
+      newArr.push(('0' + (byte & 0xFF).toString(16)).slice(-2));
+    });
+
+    return newArr.join('');
   };
 
   if (isMultiByte) {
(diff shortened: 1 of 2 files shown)


lib/m2ts/caption-stream.js
```diff
@@ -688,18 +688,32 @@ Cea708Stream.prototype.handleText = function(i, service, options) {
...
+
+  if (isMultiByte) {
+    charCodeArray = [currentByte, nextByte];
+    i++;
+  } else {
+    charCodeArray = [currentByte];
+  }
+
   // Use the TextDecoder if one was created for this service
   if (service.textDecoder_ && !isExtended) {
```
wseymour15, line 706: Would love feedback from the team here, as in the cases we are seeing issues, there is no `textDecoder_` set. I _think_ that this is a real use case and not an issue with the 708 data, but feel free to chime in if that is not the case.

dzianis-dashkevich (approved):

lib/m2ts/caption-stream.js
adrums86: I agree this is a valid use case and not a 708 data issue. In order to create a textDecoder we need an `encoding` string, which it seems is missing in this case. I think it's best to try and decode the text anyway.

adrums86 (approved):
Nice work! :shipit:

PR #445, opened 2024-09-23
feat: parse WebVTT from fmp4 segments.

adrums86:
### Add a module to parse WebVTT from fmp4 segments.
This module is based off of the ISO/IEC 14496-30 spec for carrying timed text data in ISOBMFF/FMP4 segments.
- Parses the timescale from a WebVTT init segment.
- Parses cue text, styles and timing data from WebVTT content segments.
- Exports cue objects for each parsed WebVTT cue inside the content segment with the following format:
```js
{
  cueText: 'Example cue text',
  start: 0,
  end: 1,
  settings: 'color: yellow;'
}
```
- `parseSegment` returns an array of parsed cue objects for each WebVTT segment parsed which, contains all cues from the given segment.
- Ignores both empty and invalid WebVTT cue data.


lib/mp4/webvtt-parser.js
```diff
@@ -0,0 +1,118 @@
...
+            let cueText, settings;
+
+            // contains cue text.
+            if (paylBox) {
+              cueText = textDecoder.decode(paylBox);
+            }
+
+            // settings box contains styling.
+            if (sttgBox) {
+              settings = textDecoder.decode(sttgBox);
```
dzianis-dashkevich, line 96: I think, we might want to wrap in try/catch block here

PR #457, opened 2026-09-21
fix: consume P16 parameters atomically within the CEA-708 service block

### LINKED ISSUE #456: CEA-708 P16 parameter bytes are incorrectly filtered as standalone text
### Summary

CEA-708 P16 parameter bytes are incorrectly classified as standalone G0/G1
characters by `Cea708Stream.multiByteCharacter()`. If either parameter is
outside those ranges, the parser leaves the bytes unconsumed and subsequently
interprets them as text or commands.

### Environment

- Video.js 8.24.0, bundled VHS/mux.js, HLS with in-band CEA-708 SERVICE1.
- Reproduced in Chromium and in a deterministic decoder test.
- Also present in mux.js main at e1b4d161f51782e4712d020b66d5087bd6febb3d.

### Minimal reproduction

With no explicit service encoding configured, feed the following bytes into
an initialized CEA-708 service window:

```text
18 04 10 42
```

`18` is P16, `04 10` is its two-byte operand, and `42` is following ASCII B.
Under mux.js's existing Unicode fallback, the expected text is `ÐB`
(U+0410 followed by U+0042). Currently the parameters fail
`within708TextBlock()` and are reparsed instead.

A mixed Cyrillic/Latin regression produces:

```text
 Expected: Здравей, България! English 123.
  Actual:   4@0259, J;30@8O! English 123.
```

### Proposed fix

Consume P16 atomically within the service-block/packet boundary, handle a zero
high byte as a single-byte character, and retain the existing Unicode fallback
and explicitly configured TextDecoder behavior. No new encoding assumption is
introduced; this is not a claim that every CEA-708 receiver supports Unicode P16.

We have a focused fix and regression tests ready for a PR. The patched mux.js
Node suite passes 342 tests, with 4 skipped and 0 failures, including existing
Korean/legacy-encoding fixtures. Styling and rolling-layout changes are out of
scope for this report.

XaMaB:
# fix: consume P16 parameters atomically within the CEA-708 service block

Fixes #456.

## Problem

`Cea708Stream.multiByteCharacter()` currently checks both P16 parameter bytes
with `within708TextBlock()`. That predicate classifies standalone G0/G1 text,
not bytes inside a two-byte character. When either byte fails, the parser does
not consume the parameters; subsequent iterations interpret them as ordinary
text or commands.

For example, under mux.js's **existing Unicode fallback**, P16 `18 04 10`
should reach `handleText()` as U+0410. Instead its parameters are reparsed.
A mixed test phrase `Здравей, България! English 123.` produces
`4@0259, J;30@8O! English 123.` in the unmodified Video.js 8.24.0 bundle.

## Fix

- Consume a complete P16 atomically, without G0/G1 filtering of its parameters.
- Limit consumption to the current service block and available packet data.
- Discard an incomplete P16 at that boundary instead of leaking its parameters
  into text or reading into another service.
- Handle a zero high byte as a single-byte character, avoiding a leading NUL
  when an explicit legacy decoder (for example EUC-KR) is configured.
- Retain the existing `handleText()` character mapping and configured
  `TextDecoder` path. No new character-set setting is introduced.

## Scope

This is a parsing/parameter-consumption fix. It does **not** establish Unicode
as a universal normative P16 encoding or imply support by every CEA-708
receiver. Cyrillic tests exercise the Unicode fallback already present in
mux.js. Explicit legacy encodings and existing Korean fixtures remain covered.
No changes to styling, rolling layout, timing, packetization or CEA-608.

## Tests

`npm run build-test` and `npm run test:node`: **342 passed, 4 skipped,
0 failed** on Node 18.19.1 against base commit
`e1b4d161f51782e4712d020b66d5087bd6febb3d`.

- All 304 code units U+0400–U+052F, with trailing ASCII to detect desynchronization.
- P16 operands that resemble commands.
- Zero high byte, with and without an explicit decoder.
- Truncated P16 at packet and service boundaries; following service stays intact.
- Existing CEA-608/708 and Korean/encoding regression tests.
- Separate deployed-bundle test verifies mixed Cyrillic/Latin and reproduces
  the failure on unmodified Video.js 8.24.0.
- Local HLS Chromium test discovers SERVICE1/bul, plays without playback errors,
  and decodes Cyrillic after this patch (garbled text before).

No live endpoint, credentials, customer recordings or generated player bundle
are needed in the upstream PR. Only the library and QUnit test changes belong.


XaMaB:
I noticed that #455 already addresses the same underlying P16/G0-G1 filtering issue (and references #446). Apologies for missing that existing work before opening this PR.

The overlap is intentional in effect, but there are a few differences worth considering:

- #455 retains the G0/G1 parameter filter when a TextDecoder is configured. This PR consumes complete P16 operands without that filter while retaining the existing configured-decoder path and Unicode fallback.
- This PR treats a zero high byte as a single-byte character, avoiding the leading NUL with EUC-KR. The existing Korean fixture tests still pass.
- It bounds P16 consumption to both the current service block and the available packet data. Truncated operands are not reinterpreted as text or allowed to consume the next service block.
- The added tests cover these boundaries, the zero-high-byte case, and the U+0400–U+052F range. The local Node suite passed with 342 tests passing, 4 skipped, and 0 failures.

This does not introduce a new assumption that Unicode P16 is universally supported by CEA-708 receivers; the Unicode tests exercise the fallback already present in mux.js.

Happy to consolidate with #455, contribute the additional boundary/zero-byte handling and tests there, or rebase this as a focused follow-up if that is preferable. Please let us know which approach would be most useful; there is no need to merge two overlapping fixes.

